PR #12, opened 2016-07-21
Refactor Continuous Distributions

Beakerboy:
I've moved all the distributions out of Continuous.php into separate files. I added a couple methods to the Continuous.php parent to handle the above, between, and outside CDF cases.


markrogoyski:
This has been merged. I refactored some of the code to get the unit tests working. I had to comment out the Continuous parent class' PDF/CDF functions because the parameter list differs depending on the distribution.
I see you added some new distributions as well, thanks! I'll try writing unit tests for those. Also, the new functions in Special will also need unit tests. I will update the documentation and tag a release once we have tests.

PR #31, opened 2016-07-26
Create TheilSen.php

Beakerboy:
Added a non-parametric regression technique.


markrogoyski:
Can you provide some sample data for a unit test.
Points, and the expected equation.
Thanks.

PR #59, opened 2016-08-11
Updated Inverse Function

Beakerboy:
Replacing numerical derivative with the PDF...maybe I should modify the NewtonsMethod class to accept an optional callback as the slope and just stick with that...


Beakerboy:
Thanks, That was quick...I'm enjoying a tasty beer and coding...kids are in bed and my wife is out of town.

markrogoyski:
if ($initial === null) $initial == static::median(...$params);

That == should just be = right?

Beakerboy:
yes it should. Good catch! I'll blame that on the Flanders Red.

PR #70, opened 2016-08-13
Added confidence interval and prediction intervals to regression through point.

Beakerboy:


markrogoyski:
It looks like you overwrite the x parameter with the subtraction map over the xs, and that is causing the tests to fail.

Beakerboy:
Oops, fixed

PR #99, opened 2016-08-29
Trapezoidal Rule

jakobsandberg:
I have added the Trapezoidal rule for numerical integration (quadrature) and its corresponding unit test.

If everything looks good, I can do the remaining Newton-Cotes formulas.


jakobsandberg:
I couldn't get the unit testing working on my end without manually including the TrapezoidalRule class via "include('src/NumericalAnalysis/TrapezoidalRule.php')". We could just remove that include statement (line 5 in TrapezoidalRuleTest.php).

PR #100, opened 2016-08-29
Trapezoidal Rule

jakobsandberg:
I have added the Trapezoidal rule for numerical integration (quadrature) and its corresponding unit test.

If everything looks good, I can do the remaining Newton-Cotes formulas.

Note: I fixed the two errors that were present in the previous pull request

### DIFF
--- a/src/NumericalAnalysis/TrapezoidalRule.php
+++ b/src/NumericalAnalysis/TrapezoidalRule.php
@@ -1,0 +1,129 @@
+<?php
+
+namespace Math\NumericalAnalysis;
+
+/**
+ * Trapezoidal Rule
+ *
+ * In numerical analysis, the trapezoidal rule is a technique for approximating
+ * the definite integral of a function.
+ *
+ * The trapezoidal rule belongs to the Newton-Cotes formulas, a group of methods
+ * for numerical integration which approximate the integral of a function when
+ * the antiderivative of that function is not known or is not explicity given.
+ * Instead, we use the Newton-Cotes formulas when we are given a set of inputs
+ * and their corresponding outputs for our function. We then use these values to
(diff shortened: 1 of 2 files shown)


markrogoyski:
Hi,
Thanks for contributing!

I merged the TrapezoidalRule pull request. I added some comments (the summation equation) and changed a couple variable names.

I have a question. Does the summation approximation not need to include the error term?

Thanks.

jakobsandberg:
Hi,

Looks great, I really like the comments and some of the changes you made (variable names, X and Y constants, and a shorter sorting method).

I wasn't quite sure how to include the error term. It would be very difficult (I'm not sure how) to get an accurate value for the error, as it includes terms that require the second derivative of the function we are approximating the integral for. Obviously, we can use this in the test case when we have the explicit function (which we can differentiate), but in the context of simply having points for that function, we cannot simply produce the second derivative. There ARE additional numerical techniques you can use to approximate the second derivative from (I believe) at least three points, but I'm not sure that is the best approach to take (approximating the error of an approximation via an approximation).

Let me know your thoughts on this.

markrogoyski:
I was just asking because I'm not so familiar with this function. If it is useful as is without the error term then no worries.

I noticed a bunch of online calculators instead of using points, would have the user input a function instead.
Example: https://www.easycalculation.com/integration/trapezoidal-rule.php
Example: http://www.emathhelp.net/calculators/calculus-2/trapezoidal-rule-calculator/

Would it make sense to make another version of the function (or allow the current function) to take a callable or a string equation and operate on that?

jakobsandberg:
I'm actually discussing this exact point in Issue 104 (
https://github.com/markrogoyski/math-php/issues/104) if you want to take a
look and comment there.

PR #105, opened 2016-08-30
Built abstract class NumericalIntegration, child class SimpsonsRule; updated TrapezoidalRule class; added/updated test for all three classes

jakobsandberg:
### Updates

- I built an abstract class NumericalIntegration with two children: 
 - SimpsonsRule (new)
 - TrapezoidalRule (existing).

- I moved two of the methods from the existing TrapezoidalRule class (validate() and sort()) into its parent class (NumericalIntergation) so they could also be used with the new class SimpsonsRule.

- I provided the abstract class NumericalIntegration with an abstract method approximate() that all of its children should have.

- Tests were added for the new NumericalIntegration and SimpsonsRule classes. 

- The test for the existing TrapezoidalMethod class was updated.

- Method names were changed from solve() to approximate() to better reflect the behavior of classes/methods (this was updated in the README.md examples). 

### Warnings

I was not able to generate some of the special characters in the block comments for SimpsonsRule, so the error term in the summation equation (occurs twice) needs to be updated. It's supposed to read O(h^5 f''''(x)) ("h to the fifth power times the fourth derivative of x). Instead of four apostrophes for the fourth derivative it is conventional to write (4) in small parenthesis, though I am not sure if this is possible.

### Future compatibility of callback functions

A new method should be added to NumericalIntegration to allow the approximation() methods to accept both arrays (currently compatible) and callback functions (not currently compatible), and depending on this input perform an extra step.
- If the input is an array, the approximation() methods can continue as before.
- If the input is a callback function, it should be passed into an additional method (also in NumericalIntegration) that uses the callback function along with additionally applied parameters ($start, $end, and $n) to generate an array by evaluating the given callback function at $n points between (and including ) $start and $end. Then, this array can be passed to the approximation() methods which can then proceed as they do now.

We will need to perform validation on the input callback method as well. For example, the callback function should be continuous, at least on the interval [$start, $end] to which it will be evaluated on.

### Future development of Numerical Integration techniques

There are still a number of techniques which need developing, all of which can extend the NumericalIntegration class:

- Simpson's 3/8 rule (3rd degree Closed Newton–Cotes)
- Boole's rule (4th degree Closed Newton–Cotes)
- Midpoint rule (2nd degree Open Newton–Cotes)
- Gaussian Quadrature
- Romberg Integration

It is also may be possible to extend some of these into techniques of Adaptive Quadrature and/or develop support for the approximation of multiple integrals.


markrogoyski:
Thanks for the pull request.

I updated some of the formulas, like adding the quadruple prime and raise to the 5th power characters etc.

Can you check the comment formula for h. That looks suspicious since you are computing it outside of the summation loop but the formula comment has it being computed from each summation index. It should probably be something using a and b right?

Thanks again.

jakobsandberg:
Simpsons Rule typically only allows a constant spacing between points, and thus h is constant (we do not need to compute it at each step). In this approximation() method, $a and $b denote the start and end of the interval of approximation, and so $b - $a is the total length of the interval. Thus, if there are K subintervals, the length of each subinterval (distance between each point) should be ($b - $a)/K. We get the number of subintervals by subtracting 1 from the number of points, so $K = $n - 1, where $n is the number of input points.

It would be great to generalize this rule further to allow nonconstant spacing, but there is little documentation on that online and it does not seem to be a conventional approach.

markrogoyski:
Thanks for the explanation. That confirms what I thought.
So in the comment, it says:
where h = xᵢ₊₁ - xᵢ

So that should probably say something like:
h = (x₁ - xn) / (n - 1)

jakobsandberg:
Yes, that looks more appropriate.

PR #106, opened 2016-08-30
Factorial and Hypergeometric

Beakerboy:
moved double factorial to Combinatorics
Added rising Factorial
Added Hypergeometric functions


markrogoyski:
In the generalizedHypergeometric function's equation comments there seem to be some symbols that don't render. What were these supposed to be? Maybe we can find an alternative equivalent symbol to use.

Beakerboy:
There's a bunch of Unicode symbols that don't render on my PC at work, but do on my Mac at home, like the alternate parenthesis that you use in variable names. I copied these from an online Unicode website despite them not rendering for me, thinking they would work for you. I'll double check them and change as needed.

markrogoyski:
FYI, these are two references I use a lot to get a lot of the special characters I use in equations and variable names:
http://xahlee.info/comp/unicode_math_operators.html
https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts

You have to be careful with a lot of unicode reference sites because they often display an inline graphic instead of the character because they want to be sure it displays, and copy and pasting that will often result in a square or some random character in text editor.

Beakerboy:
I'll use Wikipedia too but fileformat.info has good tables by hex code, that I can type in by hand if I'm using vim in the console.

PR #107, opened 2016-08-30
Made NumericalIntegration classes compatible with callback function inputs

jakobsandberg:
Reflected changes in comments, child classes, tests, and readme examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,16 +343,35 @@ $x        = NewtonsMethod::solve($f⟮x⟯, $args, $target, $tol, $position); //
 use Math\NumericalAnalysis\NumericalIntegration;
 
 // Trapezoidal Rule (Newton-Cotes formula)
-// Approximate the definite integral of a function given only inputs and
-// the values of the function evaluated at those inputs.
+// Approximate the definite integral of a function. Input can be either a set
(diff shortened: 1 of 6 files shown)


markrogoyski:
Can you add a unit test for this exception
https://coveralls.io/builds/7675380/source?filename=src%2FNumericalAnalysis%2FNumericalIntegration%2FNumericalIntegration.php#L114

Thanks.

markrogoyski:
How would you feel about using the ...splat operator instead of requiring the user to put the optional callback and bounds parameters in them in an array. That way the user would just add the additional parameters when using that method. It would require minimal changes, since you are already expecting an array as input.
http://php.net/manual/en/functions.arguments.php#functions.variable-arg-list

jakobsandberg:
Yeah I can add the unit test!

Also, the ...splat operator sounds like a good direction to take. So would a callback example look like this?

```
TrapezoidalRule::approximate($func, $a, $b, $c)
```

Essentially dropping the args array, allowing approximate() to have a variable number of arguments.

markrogoyski:
Exactly. It allows variable arguments. And from your function's point of view, it would get them all in an ordered array, which is what you are already expecting.

PR #115, opened 2016-09-01
Added the Bisection Method to RootFinding

jakobsandberg:
Included a test and updated the readme


markrogoyski:
Thanks for the pull request.

In the unit tests, it looks like maybe a copy and paste error; all three exception tests have a negative tolerance, so that is the exception that gets thrown and tested on all three tests, leaving the other two untested. Can you take a look into that?
https://coveralls.io/builds/7708006/source?filename=src%2FNumericalAnalysis%2FRootFinding%2FBisectionMethod.php#L84

Thanks.

markrogoyski:
I don't think this case can happen:
```
        if (!($a < $b)) {
            if ($a === $b) {
                throw new \Exception('Start point and end point of interval
                                        cannot be the same.');
```

Because the sign test above would have triggered, correct?
```
        if (Special::sgn($f⟮a⟯) === Special::sgn($f⟮b⟯)) {
            throw new \Exception('Input function has the same sign at the
                                  start and end of the interval. Choose start
                                  and end points such that the function
                                  evaluated at those points has a different
                                  sign (one positive, one negative).');
```

PR #120, opened 2016-09-01
Added RootFindingValidation trait and removed duplicate code

jakobsandberg:
I put all common validation methods inside this new trait, and then removed duplicate code from individual classes


jakobsandberg:
I didn't see you pushed some changes while I was working on this. Should I go back and pull the current branch and then do another pull request?

Also, I dropped the checks to make sure a < b because those actually unnecessary (the same result follows no matter the order you put them in) for the current techniques. But if we add any more that require them in an increasing order, we can go back and add it back in for that new technique.

PR #122, opened 2016-09-01
Added RootFindingValidation trait and removed duplicate code

jakobsandberg:
- I put all common validation methods inside this new trait, and then removed duplicate code from individual classes


jakobsandberg:
I looked into the coveralls report, and it looks like there is only one line it flagged. However, it is in LinearAlgebra/Matrix.php on line 1391 (a `break` statement) which was not affected by these commits.

markrogoyski:
Probably total lines of code went down, so the overall percentage went down. No worries. Still rounds up to 100% =)
I will eventually try to fix the Matrix one to get true 100% code coverage.

PR #129, opened 2016-09-06
Added function arithmetic

jakobsandberg:
Added capability to add and multiply functions (of one variable) to make new functions.

For example:

h(x) = f(x) + g(x)

and

h(x) = f(x) * g(x)

Note: these methods return functions themselves. This will allow for many more techniques to be constructed, such as Lagrange Interpolation, which depend on function addition and multiple to return functions themselves and not just evaluations of functions at a single point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@ Features
    - [Matrix](#linear-algebra---matrix)
    - [Vector](#linear-algebra---vector)
  * Numerical Analysis
-   - [Numerical Integration](#numerical-analysis---numerical-integration)
+   - [Interpolation](#numerical-analysis---interpolation)
+   - [Numerical Integratio](#numerical-analysis---numerical-integration)
    - [Root Finding](#numerical-analysis---root-finding)
  * Probability
      - [Combinatorics](#probability---combinatorics)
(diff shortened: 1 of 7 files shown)


markrogoyski:
Hi Jakob. Thanks for the pull request.

I'm hesitant to use any code with ```eval``` in it.

Looking at what you are doing, for example in ```add```, perhaps you can replace this:
```php
        eval('$result = function($x) use (' . $expression . ', $sum) {
            return $sum($x, ' . $expression . ');
        };');
```

with something like this:
```php
        $result = function ($x) use ($sum, $args) {
            return call_user_func_array($sum, array_merge([$x], $args));
        };
```

Try something like that and see if you can get it to work. Using ```eval``` should be a last resort.

jakobsandberg:
Mark,

Thanks for the tip! Your method worked perfectly. I'm putting in a new PR right now.

I didn't know you can pass an array to the `use` function, so I thought the only way to do this (for an arbitrary number of variables in `use()`) was via a callback function.

jakobsandberg:
Just made the new commits. Let me know if everything else looks okay.

markrogoyski:
Thanks for updating. I've future simplified it by removing some unnecessary code, and adding callable type hinting to the ...splat parameter so manual validation is not necessary.

I've merged into the develop branch. I'll merge into master when I do the next release (v0.3.0).

Thanks.

markrogoyski:
Thanks for the pull request. Merged and released in v0.3.0.

PR #131, opened 2016-09-06
Added Lagrange Polynomials with parent class Interpolation

jakobsandberg:
Included tests for both, and added Interpolation to README with Lagrange Polynomials as a feature


markrogoyski:
Thanks for the pull request. Merged and released in v0.3.0.

PR #133, opened 2016-09-07
Added Nevilles Method, slightly modified Interpolation parent class

jakobsandberg:
Included test and updated readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,20 @@ $p = Interpolation\LagrangePolynomial::interpolate($f⟮x⟯, $start, $end, $n);
 
 $p(0) // 1
 $p(3) // 16
+
+// Nevilles Method
+// More accurate than Lagrange Polynomial Interpolation given the same input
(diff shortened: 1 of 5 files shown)


markrogoyski:
Thanks for the pull request.

It looks like there is one Unicode character that is not displaying. In the NevillesMethodTest:
`// So, tol <= 24*(x-x₀)*...*(x-xₙ)/(4!) = (x-x₀)*...*(x-xₙ) where`
In the `(x-x` and then blank. I'm guessing it is supposed to be a subscript n, which generally does not display. Can you change this to maybe a regular n, which is unfortunate, but at least it will display.

Thanks.

jakobsandberg:
I replaced it with a regular n. The same unicode character was also in the `LagrangePolynomialTest.php` file, so I changed it in there as well.

markrogoyski:
Thanks. Merged to develop. Will merge to master with release v0.4.0.

PR #136, opened 2016-09-07
Added Three Point Formula with parent class Numerical Differentiation

jakobsandberg:
Included tests for both classes and updated README


markrogoyski:
Thanks for the pull request. This will go in release v0.5.0.

PR #137, opened 2016-09-07
Created cross product method

wizard97:
I noticed you didn't have a vector Cross Product method, so I created one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,7 @@ $item = $A->get(2);
 $sum    = $A->sum();
 $A⋅B    = $A->dotProduct($B);    // same as innerProduct
 $A⋅B    = $A->innerProduct($B);  // same as dotProduct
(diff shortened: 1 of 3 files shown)


markrogoyski:
Hi,
Thanks for the pull request.

Do you mind adding unit tests for the Vector cross product? You can add them to:
`tests/LinearAlgebra/VectorTest.php`

There are lots of examples in there of how to write the unit tests. Generally, the pattern is to write a generic unit test that uses a data provider, and then provide a number of different inputs with the expected output. Also, create a separate test for the Exception so there is full coverage.

Let me know if you have any questions.

Thanks!
Mark

markrogoyski:
Thanks for adding unit tests!

wizard97:
All right, I think I added the test cases you wished for. They seem to be passing!

PR #140, opened 2016-09-08
Added Polynomial class with automatic string conversion and evaluation methods

jakobsandberg:
New, I'm planning on extending numerous techniques, such as:
- Polynomial integration
- Polynomial differentiation
- Polynomial arithmetic


markrogoyski:
Thanks for the pull request and continuing to add new features. I like the __toString method making use of the exponent characters.

PR #141, opened 2016-09-08
Added integration and differentiation of polynomials

jakobsandberg:
Included tests for each, as well as a test for the fundamental theorem of calculus


markrogoyski:
I added more unit tests. I have a question. What should the return value of differentiate() be when it is run on a constant polynomial.
For example: [1]
I was expecting [0], but it seemed to have a degree of -1, which was different when I created a new polynomial of [0] input to compare against.
Can you verify how differentiate should work on a constant, and write a unit test for that case?
Thanks.

PR #149, opened 2016-09-11
Added determinant calculation for 1 x 1 matrices to pass unit test for det([1])=1

jakobsandberg:
When I added a unit test for a 1 x 1 matrix, the unit test failed, so I modified the determinant method slightly to solve this.


markrogoyski:
Thanks for pointing this out.

It looks like there are two issues: Determinant of 1x1 not being explicitly defined, and for 4x4 or larger, sometimes the sign is incorrect.

**1x1 matrix:**
I went ahead and added this in as you indicated in this pull request. I did it manually rather than pull in your changes since it looked like you were in the middle of something else and had a lot of undo commits.

**4x4 or larger matrix with incorrect sign:**
This was a bug. The reason for it is the algorithm I'm using to compute the determinant of a 4x4 or larger matrix requires first computing the reduced row echelon form of the matrix to simplify the calculation. During RREF reduction, any time a row swap occurs, you have to change the sign of the determinant. The algorithm was not checking if it was swapping a row with itself, so when that happened, it would update the number of times it swapped, even when the swap wasn't warranted when it swapped with itself.
I've fixed that with a simple check to make sure it doesn't swap a row with itself.

I've added the 5x5 unit test you tried, and several hundred more lines of unit tests just for the determinant of large matrices.

This is a good example of the importance of lots of unit tests thoroughly testing many kinds of input. This could have been caught during development with better test coverage that exposed this scenario.

It goes to show, that unit tests do not prove something correct or show that it does not have errors. Tests can only show that something has errors when the test fails. All tests passing only shows that everything you tested for works. That is why you need a lot of tests to give you confidence that it does work.

With all the additional tests, I have higher confidence that it is working properly now.

Thanks again for finding this. It was a good catch. I've merge the changes to develop. I'm closing this pull request as I did the merge and edit manually.

Thanks again!

jakobsandberg:
Mark,

Thanks for figuring this out! I had a feeling the incorrect sign was a result of RREF, but I didn't sit down and comb through the RREF function to find this error.

I'm thinking about going back to the features I've built (mostly in Numerical Analysis) and providing more rigorous unit testing. I really want to think about edge cases and extreme values to ensure the techniques work under a broad range of circumstances.

I've also been trying out the TDD methodology you outlined to me early, and I'm really starting to like it. Thanks again for explaining that!

Jakob

PR #150, opened 2016-09-12
Added two polynomial getter methods: getDegree and getCoefficient

jakobsandberg:
I slightly modified the Polynomial constructor method to remove leading 0 coefficients. This will essentially remove each leading 0 term from the coefficient array until a nonzero term is found.

For example:

```
$poly = new Polynomial([0, 0, 1]); // equivalent to p(x) = 1
$coefficients = $poly->getCoefficients() // [1], which is precisely p(x) = 1
```

Thus, at the construction of a Polynomial, I am saying `Polynomial([0, ..., 0, a, b, ..., z])` is equivalent to `Polynomial([a, b, ..., z])`. Thus, the leading 0's in the coefficient array are simply removed before the array is set as object parameter, or the degree of the object is calculated.


markrogoyski:
Thanks for the pull request.

PR #151, opened 2016-09-12
Simplifying the regularized incomplete beta function

Beakerboy:
Removed the half integer cases to use the continuous fraction.


markrogoyski:
Thanks for the pull request. It looks a lot simpler!

PR #153, opened 2016-09-13
Extending the range of gamma

Beakerboy:
Using the double factorial will extend the range of $x before it overflows.


markrogoyski:
Thanks for looking into this and updating the function. Merged into develop, with slight edit to double factorial symbol. Thanks.

PR #154, opened 2016-09-13
Added Noncentral T test and Unit Tests

Beakerboy:


markrogoyski:
Thanks for the pull request and adding another distribution!

markrogoyski:
I added comments to the distribution functions. In doing so, I had a question about the mean. Can you check what the behavior should be for ν ≤ 1? According to Wikipedia it seems like it should be undefined.
https://en.wikipedia.org/wiki/Noncentral_t-distribution#Moments_of_the_Noncentral_t-distribution

Thanks.

Beakerboy:
I think you're right. I think it reduces to a non-central Cauchy distribution when v=1. What do you think is the best way to handle an undaunted mean in this situation? The standard Sudent T will have the same issue.

markrogoyski:
I added some additional unit tests and pushed to the develop branch. Some of the tests failed. Some were slightly out of tolerance, while others got divide-by-zero errors. Can you take a look at my latest commit in the develop branch? I commented out the tests that fail and put comments as to why they are failing. My test data is based on results from this online calculator:
http://keisan.casio.com/exec/system/1180573219
Thanks.

Beakerboy:
Everything passed on the development branch. The work Indid on the Gamma function was to get these working.

markrogoyski:
I added new tests, some of which don't pass (They are currently commented out so the tests do pass). Pull in the latest changes from the develop branch and you will see them in the unit test data providers.

markrogoyski:
For something undefined, options would seem to be
* null
* \NAN (not a number)
* throw an exception

PHP seems to use \NAN for internal math functions when they are outside a defined range.

Beakerboy:
The divide by zero errors are due to the choice of a, b, and x in the test. it just so happens that the beta1 parameter in the continued fraction evaluates to 0...which is fine if the fraction continued for more parameters. I need to set the initial number of fraction to greater than 1 and it will work.

What I should also do for clarity, if remove this section into it's own helper function, and make the main regularizedIncompleteBeta more of a Factory, whereby it looks at the values of a, b, and x and determines which support function needs to be called. That way each support function can be better tested and documented.

markrogoyski:
Just curious what the state of the Noncentral T distribution is. I had some new features and thought about doing a release, but Noncentral T is also included, and it had the issue of some inputs giving errors (unit tests).

Is it OK to do a release with Noncentral T as is? If not, no rush, I can wait until it is fixed (or beta function is fixed or whatever). I like to keep momentum with lots of small continuous releases, but not at the sake or releasing something that isn't correct. Correctness is more important.

Again, not trying to rush anything. Just curious what the status is of the current code that is merged in develop. Thanks.

Beakerboy:
The divide by zero errors are fixed. The tolerance errors are indicative of larger issues. These boil down to "how accurate do we want to advertise the library?"

Many functions depend on gamma, but if you look at how it is tested, it's not very accurate. Any errors there are going to be amplified when it is used other places.

I think it would be nice to extend the base PHPUnit class to create a assertEqualsFloat method that takes in the same parameters as assertEquals, but has a static relative tolerance, like I have in the NoncentralTTest. I tested this for the entire Special Functions class and there were many existing tests that failed a 5sigma tolerance.

markrogoyski:
Was there a pull request that I missed or maybe you didn't initiate? I don't see any changes since I added the failing unit tests (commented out). They are still present in the develop branch, and if I comment any out they still fail with divide by zero. Can you check if maybe you didn't initiate a pull request? Thanks.

For the gamma function, what is it that we aren't doing that other libraries are doing that would cause variation? And can we improve it?

For unit testing floats, I can make a MathPHP UnitTest class that has the assertEqualsFloat. Would you want ε = .000001 always, or have that configurable as a parameter to the assertion?

Beakerboy:
I have no idea what other libraries do. I wrote a piece of client/server jukebox software in C++ about a decade ago, so I'm OK with the language, but the boost library is that crazy mysterious type of C++ that makes absolutely no sense to me. Then again, I was probably doing like object oriented C....but even then, there is some voodoo C out there too.

markrogoyski:
It looks like the Apache Commons Math library has the gamma family of functions, if you want to use that as another point of reference. Java, while verbose, tends to be rather straightforward.

http://commons.apache.org/proper/commons-math/javadocs/api-3.6.1/org/apache/commons/math3/special/Gamma.html

Download the source from here:
http://commons.apache.org/proper/commons-math/download_math.cgi

markrogoyski:
The C# Math.Net Numerics library might also be another point of reference to see how things are implemented. Their code seems to be pretty clean and straightforward.

Here are their Gamma functions:
https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/SpecialFunctions/Gamma.cs

PR #155, opened 2016-09-14
Added null mean when v <= 1

Beakerboy:


markrogoyski:
What is your opinion of using \NAN instead of null when the value is undefined.

For example, the built-in math functions have that behavior:
```php
sqrt(-1);
acos(2);
```
Those will return \NAN.

I tend to think that null just means a value hasn't been set. A variable exists, but it hasn't been set to anything yet.

Whereas NAN is the result of some mathematical operation that does not make sense.
Thoughts?

Beakerboy:
That sounds good.

markrogoyski:
Did you intend to close this pull request before I had pulled and merged it?

Beakerboy:
I closed because it had some issues. If I have time Sunday morning I plan on getting everything closed up as well as I can.

markrogoyski:
Oh OK. No worries. Just curious. No need to rush. Thanks for the continued contributions!

PR #162, opened 2016-09-19
Noncentral T updates

Beakerboy:
Fixed the mean when v = 1.
Fixed the divide by zero errors by calculating a minimum depth of 10 on the continued fraction.


markrogoyski:
Thanks for the pull request and continuing to improve the special functions.

Btw, if you want to include unit tests for the separate iBetaCF function that is private, a trick you can use is instead of private make it protected, which from the user point of view is the same--they can't use the function directly. And then in the unit tests create a dummy class that extends Special and create a wrapper method around iBetaCF that just calls and returns its value.

For example:
```php
class DummySpecial extends Special
{
    public static function cheating(int $m, $x, $a, $b)
    {
        return self::iBetaCF($m, $x, $a, $b);
    }
}
````

Beakerboy:
I'll keep this in mind. In this case, calling the "Master" function with particular values of x, a, and b will call this function. it's not "Isolated" but it is tested.

PR #163, opened 2016-09-19
Updated distribution means to NAN instead of null

Beakerboy:


markrogoyski:
Thanks for the pull request and updating these to NAN.

PR #165, opened 2016-09-28
Added Clamped Cubic Spline Interpolation

jakobsandberg:
Included test and updated README


markrogoyski:
Thanks for the pull request and adding another interpolation method!

It looks like there are a couple of Exceptions in the new ClamedCublicSpline class that need unit tests.

jakobsandberg:
Mark,

Thanks for catching that, I totally forgot!

I can add those in right now.

markrogoyski:
Thanks for adding the additional unit tests.

PR #166, opened 2016-09-28
Refactored Interpolation techniques to use Polynomial class instead of explicit callbacks

jakobsandberg:


markrogoyski:
Thanks for the pull request. Nice use of the classes you've been creating and building upon them.

PR #169, opened 2016-10-02
[Continuous.php] Use random_int instead of mt_rand

oittaa:


markrogoyski:
Hi. Thanks for the interest in MathPHP and thanks for the pull request.

Since `random_int` is included in PHP core since PHP 7 and MathPHP requires PHP 7, checking that the function exists would always return true.

So the issue is whether to replace `mt_rand` with `random_int`. From what I can tell, `random_int` seems to be a better random number generator, so I think it makes sense to make the change.

Can you update the pull request to just replace the function without the conditional logic? In that case, it should also return to 100% test coverage as well.

Thanks.

oittaa:
Sure! I didn't notice at first that the requirement was for PHP 7.0. Now the tests should pass.

markrogoyski:
Also, please send the pull request to the develop branch, not master. I'll merge it into develop once you make the small change I suggested in the code review.

For more information about contributing, see the contributor's guide in the project wiki. Thanks!

markrogoyski:
Thanks for making the changes. Merged into develop and will make it into the next release.

PR #176, opened 2016-12-27
Add Finance function aer for Annual Equivalent Rate.

adamrogoyski:
Similar to APY, used to compute the effective interest rate
when an APR is compounded periodically within the year.


markrogoyski:
Thank you for the pull request and continuing to add to the financial functions.

PR #180, opened 2017-01-03
Finance

adamrogoyski:
Add Finance rate function.

Update Newton's Method to handle non-convengence and infinite slopes.


adamrogoyski:
I'll update the formatting and add a function comment.

PR #197, opened 2017-03-31
Add Eigenvalues of a 2x2 matrix

Beakerboy:


markrogoyski:
Thanks for the pull request. Eigenvalues are a difficult thing to compute. Thanks for starting to work on this.

Question: What will happen if a 1x1 or 3x3 or larger is passed to this function?

aboks:
I would personally recommend to place methods for calculation of eigenvalues/eigenvectors not in the Matrix class itself, but in a separate class. As @markrogoyski already remarked, eigenvalues are difficult to compute in general, and there is no one-size-fits-all solution. Therefore I think it is important to let the consumer explicitly choose an algorithm that fits the problem/matrix at hand. By placing the logic into a separate 'algorithm'/'strategy' class (e.g. something like `ExactEigenvalueCalculator`) one forces the consumer to explicitly pick a suitable algorithm. Other algorithms (for other types of matrices) can then be added in the same way.

markrogoyski:
@aboks 
When I was thinking about this I had a similar idea. I was thinking to still put the eigenvalue method on the Matrix, but that would then make use of an Eigenvalue class that would do the computation. Based on the type of matrix the 'correct' algorithm would be called. Or, the user could specify an Eigenvalue strategy (algorithm) as an optional parameter.

I wouldn't force the user to do anything. If the strategy is left empty ```($M->eigenvalue())``` we would want it to do 'the right thing,' but also allow the user to control the behavior if necessary.

Beakerboy:
Currently, this is a simplest approach. On a larger matrix, the eigenvalues for the 2x2 matrix in the top corner would be returned. I don't know a bunch about the subject, but needed to calculate the eigenvalues of a 2x2, so this is what I came up with. There's no Exceptions thrown, or checks done. I have not tested it for cases where Algebra::quadratic fails due to complex conjugates either.

I image Algebra::cubic() can be used for a 3x3 matrix. However, row swapping or column swapping might have to happen if we want the results to exactly match Matlab.

markrogoyski:
A user will expect that if a Matrix has an eigenvalues() function that it will return the eigenvalues for the matrix, regardless of the size of the matrix. So until we have a functionally complete eigenvalues method, the Matrix class should not have an eigenvalues method.

@Beakerboy 
I understand the context of what you are doing. You need eigenvalues of a 2x2 matrix for other stuff you are working on. I do not want you to be blocked on that because a general eigenvalues method currently doesn't exist. So as a practical matter, here is my recommendation.

Create a LingerAlgebra\Eigenvalue class. Create a static method Eigenvalue::quadratic2x2 or some other appropriate name and move your pull request method in there. The only change is it will take in a Matrix as input since it is no longer a Matrix object method. Then you can use that in your other functions that require the eigenvalues of a 2x2 matrix, and there will be no confusion of what eigenvalue functionality is and isn't currently implemented.

Once a functionally complete Matrix->eigenvalues() method is implemented, the code using the quadratic2x2 method can be refactored to use the Matrix->eigenvalues() method.

PR #199, opened 2017-04-04
Eigenvalues for 2x2 or 3x3 matricies

Beakerboy:
This can also be extended to 4x4 if we implement the quartic equation (http://www.1728.org/quartic2.htm). I read somewhere that it's been proven that there is no generalized 5th power equation.


Beakerboy:
This function will throw an exception if the input is not square, or not the correct size.

markrogoyski:
Thanks for the pull request and starting the work on eigenvalues.

markrogoyski:
I didn't realize the pull request was into the master branch. I reverted the pull request and manually merged the code into develop. I also added some additional unit tests.

PR #200, opened 2017-04-04
Add Quartic Equation

Beakerboy:


markrogoyski:
Thanks for the pull request!

Two requests:
1) Can you change the destination branch from master to develop?
2) Can you add two unit test cases for when ```$cubic_roots[0]``` is 0 and ```$cubic_roots[2]``` is 0?

Also, if you are able to check for a potential imaginary number, returning \NAN would be consistent with quadratic and cubic behavior.

Thanks.

Beakerboy:
I think that we will need to get Complex Numbers working for this to be fully functional. There will be cases where the call to cubic() will return a pair of complex numbers, and these will have to be chosen to get the correct answer...even if the quartic polynomial has all real roots.

I have a ComplexNumber class started here (https://github.com/Beakerboy/math-php/blob/ComplexNumber/src/ComplexNumbers/ComplexNumber.php)

markrogoyski:
I think in this instance using a custom ComplexNumber class can work without much issue because it would only be used internally, and it sounds like you don't plan to output any ComplexNumbers--only use them to derive real roots.

As discussed in another thread, allowing ComplexNumbers to be used as user inputs and outputs to functions brings along with it the issue of how to handle all the math operations between complex and non-complex numbers. I'm not ready to redefine addition everywhere in the library. However, in cases like this where it can be used internally without affecting the API seems like a good way to introduce ComplexNumber into the library, learn, and start to consider an overall design strategy going forward of how to handle complex numbers.

Maybe someone needs to put forth a proposal to the PHP Internals Group for bidirectional operator overloading and function polymorphism for PHP 7.3. =)

markrogoyski:
Two design suggestions on the ComplexNumber class:

1) For arithmetic functions that take in a (complex) number, the code would look cleaner if ```$this``` complex number and the argument complex number were using the same thing, rather than one use an attribute (```$this->i```) and the other use a getter function (```getI()```). What you can do is create a magic ```__get``` function to get r and i. Then the arithmetic functions look like they are adding and subtracting the same things. Throw an exception if the user tries to get anything other than r and i. Don't define a ```__set``` function so r and i stay protected.

For example:
This:
```php
    $r = $this->r + $c->getR();
    $i = $this->i + $c->getI();
```
Becomes:
```php
    $r = $this->r + $c->r;
    $i = $this->i + $c->i;
```

2) For runtime safety, after the is_numeric check, instead of a blanket else, do something like ```elseif ($c instanceof ComplexNumber)```, and then throw an exception as the last else.

For example:
```php
    if (is_numeric($c)) {
        $r = $this->r - $c;
        $i = $this->i;
    } elseif ($c instanceof ComplexNumber) {
        $r = $this->r - $c->r;
        $i = $this->i - $c->i;
    } else {
        throw Exception\IncorrectTypeException('Argument must be real or complex number');
    }
```

Beakerboy:
Sounds like a plan. Do you think I should create new polynomial root functions which return complex roots, or should I add another optional argument in the function call? I like the later idea myself.

markrogoyski:
I think the optional argument is a good solution for the time being. If functions default to returning complex numbers, that might imply that you could start sending complex numbers to functions and expect reasonable results, which is not currently the case.

Beakerboy:
closing this temporarily.

PR #204, opened 2017-04-07
Complex Number Class and Complex Root of Quadratic Function

Beakerboy:
This patch adds a complex number class. This class is used in the quadratic function for cases where the roots of the polynomial are complex.


markrogoyski:
Thanks for the pull request and defining the base Complex number class.

I renamed it to MathPHP\Number\Complex. I refactored a little bit and added more unit tests.

Question: Should sqrt return an array of two complex roots?

Beakerboy:
My thought is, no, because it is operating just like the sqrt function for real numbers in that it returns the principal root.

PR #206, opened 2017-04-10
Complex roots of Cubic Polynomials

Beakerboy:
This patch extends ComplexNumber support to Cubic Polynomials. The ComplexNumber::equals() method was refactored to use an epsilon range to help floating point comparisons.


markrogoyski:
Thanks for the pull request and continuing to add complex number support.

PR #207, opened 2017-04-13
Enable square root of Complex Number with b=0

Beakerboy:


markrogoyski:
Thanks for the pull request and catching this edge case.

PR #210, opened 2017-04-14
Develop

happyjack27:
working towards adding MLE parameter estimators to the distributions.

added sample_mean and sample_variance to Probability/Distribution.php for Method of Moments and other stuff.

added some of functions to Probability/Continuous/Beta.php for MLE parameter estimation using iterative maximization of log-likelihood.

also changed a $z to a $z, which was screwing up Eclipse's syntax checker.


happyjack27:
added dirichlet and gamma distributions.

markrogoyski:
Hi,
Thank you for your interest in Math PHP and offering to contribute!

Some recommendations for this pull request:

**1) Unit tests**
Unit tests are how we gain confidence that the code is doing what it is supposed to do. Please add unit tests for all functions. Test with multiple data sets--Test common cases, edge cases, bad input, etc. 

Look at the tests/ directory to get a sense of how things are tested. Test each function thoroughly (The MathPHP testing motto is _Beyond 100% code coverage_). And, if applicable, test mathematical axioms that in turn make use of the functions. Check out the LinearAlgebra Matrix unit tests for great examples of this.

**2) Use existing code**
I noticed a couple functions that are duplicates of existing code:

- sample_mean -> Use MathPHP\Statistics\Average::mean
- sample_variance -> Use MathPHP\Statistics\Descriptive::sampleVariance

Take a look at the main MathPHP Github homepage that displays the README info. That will give you a good idea of everything that is implemented and available to you.

**3) Contributor guide**
Take a look at the contributor guides for tips
- https://github.com/markrogoyski/math-php/wiki/Contributing
- https://github.com/markrogoyski/math-php/wiki/Coding-Standards

Thanks again for offering to contribute! Let me know if you have any questions.

happyjack27:
I appreciate your attention to detail and organization, especially
regarding accidental function duplication.

One thing that struck me - and I kinda get why it was done this way - is
making everything a static function instead of instantiating a class a
distribution.

There are pros and cons to this.  Some cons are:
* the library user has to maintain their own data structures for
statistical parameters for each distribution
* innate relationships between distributions such as conjugate prior cannot
be built into the library.
* statefull things like posterior predictive distribution and bayesian
updating simply cannot be done by the library.

I guess the main difference I between stateless (static functions) and
statefull (instantiable classes).

markrogoyski:
Hi,

Not everything is a static function. Static functions are used for when the output only depends on the input.

For example, there would be no benefit to doing this:
```php
$average_object = new MathPHP\Statistics\Average;
$mean = $average_object->mean($values);
```

There is no state involved. So a static function is all you need.
```php
$mean = MathPHP\Statistics\Average::mean($values);
```

This is no different from build-in PHP functions such as cos, sin, pow, sqrt, etc.

Objects are used when there is some sort of state, and you may want to derive various different computations from that state. For example, Matrix is an object. The statistical regressions are objects. Take a look at how Matrix and statistical regressions are used and I think you will see the distinction.

For something like Bayesian updating from priors etc. it may make sense to have an object that contains state. There is no Bayesian functionality yet so that is still to be defined.

happyjack27:
Yeah sorry I should have specified I was just talking about
Probability/Distributions/ .

Also thinking about down the road with more complicated things that require
state - in that folder in particular - so then you might have a static
function AND a member function that both so the same thing - or worse, a
instantiable class that doesn't implement everything the static functions
do, and vice versa, so you have to mix and match.

At that point, a major refactoring of static functions into member
functions could break backwards compatibility.  While leaving both might
be confusing to the user.

Beakerboy:
If you write up a little bit on how you would imagine an instantiable Distribution, we might be able to figure something out. Would it be possible to create a general instantiable class which contains one of the static classes, and member functions which, for example, relate to the estimation of the member distribution parameters? I'm thinking about a strategy similar to how I have the least squares Object designed to refer to a static "model".

happyjack27:
I have a few ideas.  I'll check out the least squares object.

Obviously it's a tough problem.   Just trying to head off a future Sophie's
choice.  (Sun Tzu says: "Deal with large problems when they are small.")

happyjack27:
apache commons math has a nice implementation of probability distributions
for java:

http://commons.apache.org/proper/commons-math/download_math.cgi

that might serve as a good model, though it's pretty basic. (e.g. no
parameter estimators)

i've decided that for what i'm working on now, i'll just use java (and
apache commons math) instead of php.

markrogoyski:
Closing this pull request as the Gamma and Dirichlet distributions have been implemented independent of this pull request in v0.30.0.
Please don't hesitate to create a new pull request for parameter estimation/average log likelihood functions.
Thanks again for your interest in MathPHP.

PR #213, opened 2017-04-19
Add unitCircle method to Trigonometry Class

Beakerboy:


markrogoyski:
Thanks for the pull request.

Question: Do you think there would eventually be other classes in the Trig namespace? For example, Angle, Triangle, maybe even UnitCircle as a class?

Maybe this will eventually grow into something like:
MathPHP\Trigonometry\UnitCircle::points($n)
MathPHP\Trigonometry\Angle
MathPHP\Trigonometry\Triangle
...

Or maybe not. I'm not sure.

Beakerboy:
I have no plans to add anything.

PR #214, opened 2017-04-19
Refactor method

Beakerboy:
Simplified function calls


Beakerboy:
oops, this was to master

PR #215, opened 2017-04-19
Refactor Continuous Inverse Method

Beakerboy:
Simplified function calls


markrogoyski:
Thanks for the pull request. Good change.

PR #218, opened 2017-04-25
Polynomial difference

Beakerboy:
Added a function to take the difference of two polynomials.


markrogoyski:
Thanks for the pull request and adding subtract and providing tests.

PR #219, opened 2017-04-27
Added Polynomial::scalerMultiply() and Polynomial::roots()

Beakerboy:


markrogoyski:
Thanks for the pull request and nice unit tests.

src/Functions/Polynomial.php
```diff
@@ -376,4 +378,40 @@ public function multiply(Polynomial $polynomial): Polynomial
...
+        switch ($this->degree) {
+            case 1:
+                return [-1 * $this->coefficients[1] / $this->coefficients[0]];
+            case 2:
+                return Algebra::quadratic(...$this->coefficients);
+            case 3:
+                return Algebra::cubic(...$this->coefficients);
+            case 4:
+                return Algebra::quartic(...$this->coefficients);
+            default:
```
markrogoyski: What should happen for a zero-degree polynomial (constant function)? Should that return empty array instead of defaulting to [\NAN]?

src/Functions/Polynomial.php
```diff
@@ -376,4 +378,40 @@ public function multiply(Polynomial $polynomial): Polynomial
...
     }
+
+    /**
+     * Multiply a polynomial by a scalar value
+     *
+     * @param number $scaler value we are multiplying with our current polynomial
+     *
+     * @return object The product of our polynomial object and the scaler, also a polynomial object
+     */
+    public function scalerMultiply($scaler): Polynomial
```
markrogoyski: If you don't add a type declaration here you could pass in anything and try to multiply by it and get an error. I think adding a float type declaration is better than checking for is_numeric within the function..

src/Functions/Polynomial.php
markrogoyski: What would roots return for a zero-degree polynomial with constant value 0? A generator that yields every number from -infinity to +infinity (represented by \PHP_INT_MIN to \PHP_INT_MAX)?

Beakerboy:
I thought about those edge conditions and figured that NAN was the simplest solution.  

I'll add the type hint shortly.

markrogoyski:
For the roots default case where the degree is > 4, is there a clean way to use one of the NumericalAnalysis\RootFinding methods to find all the roots? Or at least provide some of them?

Beakerboy:
I was thinking that I could use the Descartes Sign Rule to test how many positive, negative, and complex roots there are, and then use a numerical method to find one. Then polynomial division can be added to the class to reduce the order by one, and the operation can repeat. Simple cases like when it has a root equal to zero, if the highest order is zero could be added now.

markrogoyski:
Do you want to create a new pull request for the higher-degree polynomial roots? Or should I wait to merge this for that to be developed?

Beakerboy:
If it's up to me, I'd say push these if you approve, and we can work on that later. I have a plan where I'm building on this to make an ObjectMatrix next, which will allow Polynomial and Complex matricies. A polynomial Matrix will simplify Eigenvalue calculations, while the Complex Matrix will fulfill a feature request.

Check out how simple the eigenvalue calculation is compared to the current one, and this will work for all sizes. The nested for loops could be replaced by some sort of "convert a numeric matrix to a Polynomial Matrix" method, and then subtract a diagonal matrix of lambda.
````
// Convert the numerical matrix into an ObjectMatrix
        $B_array = [];
        for ($i = 0; $i < $m; $i++) {
            for ($j = 0; $j < $m; $j++) {
                if ($i == $j) {
                    // Subtract λ from each of the diagonal elements.
                    $polynomial = new Polynomial([-1, $A[$i][$j]], 'λ');
                } else {
                    $polynomial = new Polynomial([$A[$i][$j]], 'λ');
                }
                $B_array[$i][$j] = $polynomial;
            }
        }
        
        $B = MatrixFactory::create($B_array);
        $det = $B->det();
        
        // Calculate the roots of the determinant and convert into a diagonal matrix
        $eigenvalues = MatrixFactory::create($det->roots());
        return $eigenvalues;
````

markrogoyski:
Thanks for the info. I look forward to that pull request!

Since Polynomial is still an internal-use class at this point, it is OK if the functionality isn't complete. I'll create an issue to implement this functionality so it does not get forgotten. I'd like functionality to match user expectations when something gets published.

PR #222, opened 2017-05-01
Refactor EIgenvalue method.

Beakerboy:


src/LinearAlgebra/ObjectSquareMatrix.php
```diff
@@ -0,0 +1,114 @@
...
+class ObjectSquareMatrix extends SquareMatrix
+{
+    /**
+     * Check that the matricies are the same size
+     *
+     * @throws MatrixException if matrices have a different number of rows or columns
+     */
+    private function checkEqualSizes(Matrix $B)
+    {
+        if ($B->getM() !== $this->m) {
```
markrogoyski: Do you also need to check n (number of columns)? I think the argument Matrix could have the same number of rows but not be square.

markrogoyski:
There are a lot of matrix methods that seem like they will fail if called on the ObjectSquareMatrix because built-in arithmetic operators are used on matrix elements. Is the intention that ObjectSquareMatrix only be an 'internal' class used to calculate other things, and not something a user would use on their own? If that is the case, then maybe it would make sense to have some sort of Internal namespace to indicate it is not part of the public API.

src/LinearAlgebra/ObjectSquareMatrix.php
Beakerboy: Good point. I should check that the other is square, or that the n's are equal.

Beakerboy:
Correct, in the long run, any Matrix method that has an arithmetic operator will need to be rewritten. Also, many method WILL fail depending on the type of object. For example, polynomial division will rarely return another Polynomial, so any matrix method that divides two elements will return \NAN. 

In the long run, this class will be useful for matricies of Complex numbers, so it will be part of the API eventually. As of right now, it just makes the EIgenvalue method much simpler, and independent of the matrix size.

Also, here we will see a consequence of type hinting with a SquareMatrix. This class must be an extension of SquareMatrix due to type hinting on the minorMatrix function in the Matrix class. Any function in this class will have to be duplicated line-for-line in a non-square version of this class. Would it make sense to just have all the matrix children extend Matrix, and have the square versions use a trait that contains methods involving square shaped matrices? We would then check for isSquare instead of type hinting. I feel like we may have had this conversation before though...

markrogoyski:
Thanks for the explanation.

For the SquareMatrix, I see your point. Even though as a precondition of the minorMatrix function is that the matrix being worked on is square itself (checked vis the isSquare method), the non Square version permits Square-only functionality provided that the isSquare method passes. If you had a generic ObjectMatrix that extended Matrix, in order to return a SquareObjectMatrix, that has to extend SquareMatrix and not Matrix. This is a case of single inheritance being limiting.

I can see how this will become an issue. I can't immediately think of a solution. I'll give it a little more thought, but I think you are right. If we don't want to duplicate code, all matrices should probably extend from Matrix, and removing the SquareMatrix type hint may make some things easier.

I like it when the language does automated precondition/postcondition checks for you. But we should not allow the language to get in the way of good design.

Beakerboy:
It would probably be a good idea to ensure that all the elements in the newly created matrix are objects of the same type.
I propose we add a property to this class that tells what kind of objects are in the matrix. 
Next, override the constructor to check the instanceof on the elements within the for loop.
Should we allow a mixture of objects and numeric data, or force all objects? We could add a Object::castNumeric(number) method to standardize casting constants (for Polynomials) or real numbers (Complex) to the corresponding objects.

This way the arithmetic operations can check that the two ObjectMatrix objects are compatible.

markrogoyski:
That sounds like a good direction to take it. It is definitely a good idea to validate that the data makes sense.

It seems like these are the OK cases:
- All objects of the same type
- All objects of the same type or numeric types
- All numeric types (In this case, do you cast to Polynomial or Complex?)

Beakerboy:
For the final bullet point, we would probably want a method where the user specifies an object to cast the numeric into. We could develop an Interface, and ensure the class implements it. Then you would be confident that any user defined classes would work correctly.

Beakerboy:
I'm closing this until I add the new ideas

Beakerboy:
I think it would be easiest to just ensure that the constructor can understand what to do if it just receives a single numeric value, instead of creating a castNumeric() method . In the Complex Number, I can just make the imaginary argument optional. For the Polynomial, I can remove the array type hint, and add a check, for if it's an array, or if it's numeric.

Any suggestion?

markrogoyski:
I like the idea of just having the constructor handle any logic necessary for creating a usable object, rather than having separate methods after the object is constructed. The object should be in a usable state once it is constructed.

I don't think Polynomial needs to change. If you have a number, just instantiate it like this:
```php
$x = 5;
$poly = new Polynomial([$x]);
```

Instead of creating Complex numbers to represent real numbers, you could create an Integer or Real number class that is derived from a Complex number, which itself would always have i being equal to 0. But would that then complicate the return value logic if the Real is on the left hand side of arithmetic with a Complex number?
For example:
```php
$complex2 = $real->add($complex);
```
In that case, the class would need to know what to return. Otherwise, you might end up with something silly like a Real number with an imaginary part.

The easiest solution probably is just to use Complex numbers internally with i set to 0.

PR #224, opened 2017-05-07
Add an Interface

Beakerboy:
I will use this to ensure that any object used in a Object matrix has the necessary methods to work correctly.

### DIFF
--- a/src/Number/MatrixObjectInterface.php
+++ b/src/Number/MatrixObjectInterface.php
@@ -1,0 +1,9 @@
+<?php
+namespace MathPHP\Number;
+
+interface MatrixObjectInterface
+{
+    public function add($object);
+    public function subtract($object);
+    public function multiply($object);
+}
--- a/src/Number/Complex.php
+++ b/src/Number/Complex.php
@@ -13,7 +13,7 @@ use MathPHP\Functions\Special;
  * part of the complex number.
  * https://en.wikipedia.org/wiki/Complex_number
  */
-class Complex
+class Complex implements MatrixObjectInterface
 {
     /**
      * Real part of the complex number


src/Number/MatrixObjectInterface.php
```diff
@@ -0,0 +1,9 @@
+<?php
+namespace MathPHP\Number;
+
+interface MatrixObjectInterface
```
markrogoyski: How about renaming this something like:
- MatrixObjectOperations
- MatrixObjectArithmetic

src/Number/MatrixObjectInterface.php
```diff
@@ -0,0 +1,9 @@
...
+{
+    public function add($object);
```
markrogoyski: The object parameters should also implement the interface, correct? If so, add the interface type declaration to the parameter.

markrogoyski:
Thanks for the pull request. This makes sense to have objects implement the interface and then you can use the interface type declaration to ensure compatibility.

src/Number/MatrixObjectInterface.php
Beakerboy: You choose it and I'll change it.

src/Number/MatrixObjectInterface.php
Beakerboy: Maybe I should rename it $object_or_scaler.

src/Number/MatrixObjectInterface.php
markrogoyski: Eventually I imagine there may be more than add/multiply/subtract, so how about MatrixObjectOperations?

src/Number/MatrixObjectInterface.php
markrogoyski: OK. In that case, when you add the PHP docBlock comments, you can add like:
```php
@param mixed $object_or_scalar Explanation goes here
```

src/Number/MatrixObjectInterface.php
markrogoyski: Actually, isn't this interface going to determine if the objects themselves know how to handle arithmetic operations with other objects and/or scalars? The fact that these objects will be used in a matrix might be just a detail. Are there any other contexts where these objects might get used. And if yes, maybe a more general name, like ObjectOperations or ObjectArithmetic.

Beakerboy: That's an excellent point. ObjectArithmetic it is.

PR #225, opened 2017-05-09
Polynomial implements ObjectArithmetic 

Beakerboy:


markrogoyski:
Thanks for the pull request and integrating the new ObjectArithmetic interface.

src/Functions/Polynomial.php
```diff
@@ -58,12 +60,19 @@ class Polynomial
...
      * that polynomial as its object parameters.
      *
-     * @param array $coefficients An array of coefficients in decreasing powers.
+     * @param mixed $coefficients An array of coefficients in decreasing powers
+     *                            or a numeric value.
      *                            Example: new Polynomial([1, 2, 3]) will create
      *                            a polynomial that looks like x² + 2x + 3.
      */
-    public function __construct(array $coefficients, $variable = "x")
+    public function __construct($coefficients, $variable = "x")
```
markrogoyski: I think I'd leave the array type declaration and keep Polynomial construction consistent. If it is a zero-degree polynomial it will be instantiated just like any other degree, similar to how you do it in checkNumericOrPolynomial:
```php
$scalar = 5;
$polynomial = new Polynomial([$scalar]);
```
Is there a particular reason to allow scalar types without being in an array?

Beakerboy: I thought there was a reason for that, but it escapes me now. The Complex object is constructed with two numeric values. For consistency, do you think that should be changed to an array?

markrogoyski:
The Complex object's properties are fixed (real and imaginary), so it makes sense to pass them in individually, whereas the Polynomial could have any number of coefficients passed to it.

Something to consider with the Polynomial is to keep the constructor as an array, or use a ...variadic parameter list. With the variadic, you could use a type declaration. With an array, you have to separately check that every element is numeric prior to assigning the values from the array. The user might already have the elements in an array so an array as input might be easier. Depends on how it will be used.

In a random note, it seems like Github removes comments that are 'addressed' by a new commit. I'm not sure how to respond directly to the comment after it was addressed.

Beakerboy:
I can still see the old comments by clicking "Show Outdated", and it looks like I can still comment on it too.

tests/Functions/PolynomialTest.php
```diff
@@ -804,4 +858,12 @@ public function dataProviderForRootsNAN()
             ],
         ];
     }
+
+    public function testException()
+    {
+        $this->setExpectedException('MathPHP\Exception\IncorrectTypeException');
```
markrogoyski: Going forward, use ```$this->expectException``` instead of the deprecated ```$this->setExpectedException``` and use class constants instead of strings ```Exception\IncorrectTypeException::class```.

PR #227, opened 2017-05-10
Created ObjectSquareMatrix

Beakerboy:
With this class, matrix addition, subtraction, and determinants can be performed on objects.

### DIFF
--- a/src/LinearAlgebra/ObjectSquarematrix.php
+++ b/src/LinearAlgebra/ObjectSquarematrix.php
@@ -1,0 +1,172 @@
+<?php
+
+namespace MathPHP\LinearAlgebra;
+
+use MathPHP\Exception;
+
+/**
+ * ObjectSquareMatrix
+ *
+ * The objectSquareMatrix extends Matrix functions to a matrix of objects.
+ * The object must implement the MatrixArithmetic interface to prove
+ * compatiblity. It extends the SquareMatrix in order to use Matrix::minor().
+ */
+class ObjectSquareMatrix extends SquareMatrix
+{
+    /*
+     * The type of object that is being stored in this Matrix
+     * @var string
+     */
+    protected $object_type;
+
+    /*
+     * Constructor
+     *
+     * The constuctor follows performs all the same checks as the parent, but also checks that
+     * all of the elements in the arry are of the same data type.
+     * @param array of arrays $A m x n matrix
+     *
+     * @throws BadDataExpcetion if any rows have a different column count
+     * @throws IncorrectTypeException if all elements are not the same class
+     * @throws IncorrectTypeException if The class does not implement the ObjectArithmetic interface 
+     */
+    public function __construct(array $A)
+    {
+        parent::__construct($A);
+
+        if (in_array("MathPHP\Number\ObjectArithmetic", class_implements($A[0][0]))) {
+            $this->object_type = get_class($A[0][0]);
+        } else {
+            throw new Exception\IncorrectTypeException("The object must implement the interface.");
+        }
+        foreach ($A as $i => $row) {
+            foreach ($row as $object) {
+                if (get_class($object) != $this->object_type) {
+                    throw new Exception\IncorrectTypeExpcetion("All elements in the matrix must be of the same type.");
+                }
+            }
+        }
+    }
+
+    /**
+     * Get the type of objects that are stored in the matrix
+     *
+     * @return string The class of the objects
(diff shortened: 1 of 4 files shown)


Beakerboy:
I have not implemented Matrix multiplication yet. The parent allows a matrix to be multiplied by a Vector. if you wish to maintain that, we will have to perform the Vector::dotProduct() algorithm within the ObjectMatrix::multiply() method to avoid creating an ObjectVector class. Thankfully, it looks like the Vector class will not complain if it is constructed with objects. Many of the native method would fail though.

src/LinearAlgebra/ObjectSquarematrix.php
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @throws BadDataExpcetion if any rows have a different column count
+     * @throws IncorrectTypeException if all elements are not the same class
+     * @throws IncorrectTypeException if The class does not implement the ObjectArithmetic interface 
+     */
+    public function __construct(array $A)
+    {
+        parent::__construct($A);
+
+        if (in_array("MathPHP\Number\ObjectArithmetic", class_implements($A[0][0]))) {
```
markrogoyski: You can use ```instanceof``` to check if an object implements an interface. It seems cleaner than matching a string against an array.

```php
if ($A[0][0] instanceof MathPHP\Number\ObjectArithmetic) {
   ...
}

File: src/LinearAlgebra/ObjectSquarematrix.php [hunk 2]
```diff
@@ -0,0 +1,172 @@
+<?php
```
markrogoyski: PSR autoloading and style guides say the file name and class name have to match. matrix is not CamelCased in the file name.
ObjectSquarematrix -> ObjectSquareMatrix

tests/LinearAlgebra/ObjectSquareMatrixTest.php
```diff
@@ -0,0 +1,197 @@
...
+        $A = MatrixFactory::create([[$object]]);
+    }
+
+    public function testMatrixAddException()
+    {
+        $polynomial = new Polynomial([1, 4, 7]);
+        $complex = new Complex(1, 4);
+        $A = MatrixFactory::create([[$polynomial]]);
+        $B = MatrixFactory::create([[$complex]]);
+        $this->expectException(Exception\IncorrectTypeException::class);
```
markrogoyski: Thanks for using the more modern format.

src/LinearAlgebra/ObjectSquarematrix.php
```diff
@@ -0,0 +1,172 @@
...
+    {
+        if ($B->getM() !== $this->m || $B->getN() !== $this->n) {
+            throw new \MathPHP\Exception\MatrixException('Matrices are different sizes');
+        }
+        if ($B->getObjectType() !== $this->object_type) {
+            throw new \MathPHP\Exception\IncorrectTypeException('Matrices must contain the same object types');
+        }
+    }
+
+    /*
```
markrogoyski: Minor thing: docBlock starts with ```/**```
If you use an IDE or SublimeText or something with syntax highlighting, usually the @tags will not syntax highlight properly without the second * so it is a visual clue that something is not right. Github does the same.

markrogoyski:
Thanks for the pull request. Looks good. Nice tests. Just a couple of minor improvements to make.

Btw, we talked about removing the SquareMatrix. I'm fine with that. Doesn't need to be done now in this pull request. I imagine a lot of tests will break. If it is still something you think we should do let me know and I'll put it on my backlog of things to do.

src/LinearAlgebra/ObjectSquarematrix.php
Beakerboy: Good call, I didn't know this worked on interfaces

Beakerboy:
Yeah, I think reorganizing the inheritance of the Matrix classes will be helpful in the long term. We can use traits to manage multiple-inheritance, and just have everything inherit the main Matrix class. If you have a "Backlog", can you use the GitHub Projects manager to list them?

markrogoyski:
I'll check out the Projects manager. Thanks.

markrogoyski:
I created a development backlog project: https://github.com/markrogoyski/math-php/projects/1

Beakerboy:
I see you have "covariant matrix" on the list. The Matrix class already has a method to calculate the covariance matrix.

PR #228, opened 2017-05-11
Eigenvalue Class Changes

Beakerboy:
Changed the method name.
Refactored the method to make it more generic
Added Eigenvector method

### DIFF
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -3,64 +3,107 @@ namespace MathPHP\LinearAlgebra;
 
 use MathPHP\Algebra;
 use MathPHP\Exception;
+use MathPHP\Functions\Polynomial;
+use MathPHP\Functions\Map\Single;
+use MathPHP\Functions\Special;
 
 class Eigenvalue
 {
     /**
-     * Produces the Eigenvalues for a 2x2 or 3x3 matrix
+     * Produces the Eigenvalues for square 2x2 - 4x4 matricies
      *
      * Given a matrix
      *      [a b]
      * A =  [c d]
      *
-     * Find λ₁ and λ₂ such that the determinant of:
-     *      [a-λ, b   ]
-     *      [c,    d-λ] = 0
+     * Find all λ such that:
+     *      |A-Iλ| = 0
      *
-     * or ad - λ(a+d) + λ² - cb = 0
+     * This is accomplished by finding the roots of the polyniomial that
+     * is produced when computing the determinant of the matrix. The determinant
+     * polynomial is calculated using polynomial arithmetic.
      *
      * @param Matrix $A
      *
-     * @return DiagonalMatrix of eigenvalues
+     * @return array of eigenvalues
      *
      * @throws Exception\BadDataException if the matrix is not square
-     * @throws Exception\BadDataException if the matrix is not 2x2 or 3x3
+     * @throws Exception\BadDataException if the matrix is not 2x2, 3x3, or 4x4
      */
-    public static function quadratic(Matrix $A): DiagonalMatrix
+    public static function eigenvalue(Matrix $A): array
     {
         if (!$A->isSquare()) {
             throw new Exception\BadDataException('Matrix must be square');
         }
 
         $m = $A->getM();
-        if ($m < 2 || $m > 3) {
-            throw new Exception\BadDataException("Matrix must be 2x2 or 3x3. $m x $m given");
+        if ($m < 2 || $m > 4) {
+            throw new Exception\BadDataException("Matrix must be 2x2, 3x3, or 4x4. $m x $m given");
         }
-
-        $A = $A->getMatrix();
-
-        if ($m === 2) {
-            $a = -1;
-            $b = $A[0][0] + $A[1][1];
-            $c = $A[1][0] * $A[0][1] - $A[0][0] * $A[1][1];
-            $eigenvalues = Algebra::quadratic($a, $b, $c);
-        } else {
-            $a  = $A[0][0];
-            $b  = $A[0][1];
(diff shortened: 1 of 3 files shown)


Beakerboy:
The eigenvalue method now returns the values as an array instead of a diagonal matrix. Should I sort them largest to smallest?

markrogoyski:
For all the root functions, those weren't sorted, but it seems to make
sense to sort those. Unit tests right now are highly dependent on the
implementation and the order they come back. If you change the
implementation, the unit tests might break even though all the same roots
come back. I don't see any issue with sorting. Lowest to highest or highest
to lowest though? =)

Beakerboy:
Just look this over and merge as-is if you like it. I was just asking in case you knew of a standardized way of presenting the results. Matlab returns unit eigenvectors, which I do as well, but even that is just because of convention.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -3,64 +3,116 @@
...
      *
-     * @return DiagonalMatrix of eigenvalues
+     * @return array of eigenvalues
      *
      * @throws Exception\BadDataException if the matrix is not square
-     * @throws Exception\BadDataException if the matrix is not 2x2 or 3x3
+     * @throws Exception\BadDataException if the matrix is not 2x2, 3x3, or 4x4
      */
-    public static function quadratic(Matrix $A): DiagonalMatrix
+    public static function eigenvalue(Matrix $A): array
```
markrogoyski: The intention of this class is to provide various methods for computing eigenvalues, so each method (only one so far that you have written) will be named after the algorithm used to compute them. Previously it was using the quadratic roots so it was named after that. Now it is doing some sort of Polynomial determinant magic--does this algorithm have a name? What should the new name of this method be?

tests/LinearAlgebra/EigenvalueTest.php
```diff
@@ -1,8 +1,6 @@
 <?php
-namespace MathPHP\Tests\LinearAlgebra;
+namespace MathPHP\LinearAlgebra;
```
markrogoyski: The unit tests were all moved to the ```MathPHP\Tests``` namespace. There were two reasons:

1) I started using a common MatrixDataProvider class in multiple test files. In order to get autoloading to work in the unit tests they needed their own namespace.

2) I like the idea that tests reflect how code might actually be used. Most likely someone who uses this library is not going to be using the MathPHP namespace--they will be in their own namespace or none at all possibly. So by having the tests also be in their own namespace separate from the core code, the tests are more like a real-world usage of the code.

markrogoyski:
Thanks for the pull request and continuing to add to and improve the eigenvalue methods. Just a couple minor comments.

src/LinearAlgebra/Eigenvalue.php
Beakerboy: The algorithm does not have a name because it's just using the definition of the EIgenvalue. I'll explain it more in case the comments in the code are not clear enough. 

If A is a matrix, x is a column vector, and λ is a number:
The EIgenvalue is defined as the constant 
such that AX=λX. If you do some matrix magic, this means that (A-λI)X=0 or  |A-λI| =0. The "polynomial magic" is just doing the algebra of calculating this determinant with λ in it, collecting all the terms, and since it's set to zero, finding the roots. In essence, this method is just a polynomial rootfinding method. If we extend our ability to find polynomial roots somehow, this algorithm will be extended.

I have no idea what non-rootfinding techniques are used to find eigenvalues.

The eigenvector algorithm is a little more magical. Since the determinant of A-λI is zero then it is not invertible. However, since the set of Eigenvector solutions are all linearly scaled from one another we can set the value of any one unknown and find the  minor matrix with an a which is invertible so we can find the solution for the remaining unknowns. 

While working on this! I guess the one thing I see I am missing is how to handle the case where no minor matrix is found. I will look to see how Matlab handles this case, for example what the eigenvalues and eigenvectors are for a square matrix of all zeros.

Beakerboy: Let me know if you have any name suggestions. Eigenvalue::closedFormPolynomialRoots() is probably the most accurate.

Beakerboy: It looks like the Eigenvector method can be improved by using the reduced row echelon form of *A*-λ*I*. I'll have to look at this some more. Improvements will be forthcoming in a future pull request.

markrogoyski: Thanks for the explanation. Really cool.
Here is a list of eigenvalue algorithms: https://en.wikipedia.org/wiki/Eigenvalue_algorithm
A lot of them may only find a single value, or only apply to a certain type of matrix. Maybe your approach is innovative and unique? Perhaps you can name it =)

Is there a particular reason why it would only apply to 2x2 to 4x4 matrices? Or is that just because the Polynomial class right now only supports up to 4 degrees?

Beakerboy: The [Abel–Ruffini theorem](https://en.wikipedia.org/wiki/Abel%E2%80%93Ruffini_theorem) says that closed form solutions for polynomials only exist for polynomials with order less than 5. I doubt it's unique, it's just like [The standard 2x2 algorithim](https://en.wikipedia.org/wiki/Eigenvalue_algorithm#2.C3.972_matrices) but up to the fourth order using cubic and quintic root solving.

Beakerboy:
I have a more robust process in the pipeline, but the Matrix::rref() bug is preventing one of the unit tests from passing.

markrogoyski:
Sorry that the rref fix is blocking you. I'll let you know when that is fixed.

markrogoyski:
Please pull from the latest develop branch and see if the RREF fixes your issue.

PR #229, opened 2017-05-15
Refactored Polynomial::add() method to be simpler and faster

jakobsandberg:
What are your thoughts on this refactor? I'd like to apply a similar pattern to other methods in this class to clean things up (and keep consistency).

I added a test to make sure logic worked when instance polynomial has a smaller degree than the argument polynomial. The old tests only account for situations when the degrees were equal OR the instance polynomial has a **larger** degree than the argument polynomial.


Beakerboy:
Hoh can replace your array_map with Map\Multi::add(). Otherwise, yeah, pretending zeroes to the shorter array in order to add then seems simpler then flipping both arrays, then adding.

Beakerboy:
It looks like this change was not based on the current version of this file. The class has changed somewhat since you last edited it. You might want to update the branch and reapply these changes. There is also Polynomial::subtract() which would likely benefit from the same modification.

jakobsandberg:
@Beakerboy Just updated the PR. I was using `master` as a base but I just rebased to `develop`.

markrogoyski:
Thanks for the pull request. I agree this makes the add method easier to understand.

PR #231, opened 2017-05-20
Failing unit test

Beakerboy:
There is a bug somewhere in the matrix code. Some code I am working on failing because this Matrix::rref() is incorrect. It is currently returning an identity matrix instead of:

    [1 0 2]
    [0 1 1]
    [0 0 0]

### DIFF
--- a/tests/LinearAlgebra/MatrixDecompositionsTest.php
+++ b/tests/LinearAlgebra/MatrixDecompositionsTest.php
@@ -313,6 +313,18 @@ class MatrixDecompositionsTest extends \PHPUnit_Framework_TestCase
         return [
             [
                 [
+                    [3, -4, 2],
+                    [-2, 6, 2],
+                    [4, 2, 10],
+                ],
+                [
+                    [1, 0, 2],
+                    [0, 1, 1],
+                    [0, 0, 0],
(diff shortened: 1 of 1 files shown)


markrogoyski:
Thanks for the unit test to validate this. I have also run into this and know about the issue and have a different test case that also reproduces this. This was actually the next thing on my list of things to do.
My plan is to change rref to start by using the new ref function and then just continue from there, rather than use a different algorithm which I'm pretty sure at this point doesn't work. I added many additional unit tests to the new ref algorithm so using that as the starting point for finishing the reduction to rref I think is the way to go.
I'll let you know when I implement the fix.
Thanks again for the additional test case.

markrogoyski:
Added this test case manually to my refactored code. No need to merge the pull request.

I changed the algorithm for RREF. Your test case, as well as a couple of mine, which failed on the old algorithm, now pass on the new algorithm. There are no regressions introduced. I think this fixes the issue. Pull from develop to get the code with the fix.

If you notice any other issues don't hesitate to let me know.

One thing I need to consider is a way to have a scaling measure of what is considered zero. But for the meantime, this simplistic approach seems to work for most values.

PR #232, opened 2017-05-23
Refactored Eigenvalue and added Eigenvector method

Beakerboy:
This is a much improved eigenvector method from what I had previously. Now, when the same eigenvalue is present multiple times, the two (or more) eigenvectors are orthogonal. There were cases where the previous algorithm wouldn't work, and this covers them.

### DIFF
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -91,7 +91,7 @@ class Eigenvalue
      */
     public static function eigenvector(Matrix $A, array $eigenvalues = []): Matrix
     {
-        if ($eigenvalues = []) {
+        if ($eigenvalues == []) {
             $eigenvalues = self::closedFormPolynomialRootMethod($A);
         }
         if (!$A->isSquare()) {
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -89,15 +89,18 @@ class Eigenvalue
      *
      * @return Matrix of eigenvectors
      */
-    public static function eigenvector(Matrix $A, array $eigenvalues): Matrix
+    public static function eigenvector(Matrix $A, array $eigenvalues = []): Matrix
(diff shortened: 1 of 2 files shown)


markrogoyski:
Thanks for the pull request and continuing to improve the eigenvalue/vector methods.

Did you decide on a name for this method? Perhaps ```closedFormPolynomialRootMethod``` as you suggested earlier. 

The eigenvector method also has the constraint of 2x2 - 4x4 matrices, correct? Add a comment in the method's docBlock and rename to closedFormPolynomialRoots or something appropriate and the pull request will be good to merge.

Thanks.

Beakerboy:
Will do. In theory, this eigenvector method would work for any sized matrix, if the proper EIgenvalue array could be provided. For example, if the Polynomial root function were to use Newton's method to find one root, reduce the equation to a quartic, then use the closed-form method to find the rest, the eigenvector function would work as-is. Alternatively, if the EIgenvalue method accepted an array of eigenvalues as an input, it would work for any sized matrix. Would you prefer this?

markrogoyski:
I like that the eigenvector method could work on a Matrix without needing an array of already-computed eigenvalues. Perhaps as an option, provide the array of eigenvalues, and if not provided, the default behavior will be to compute the eigenvalues. That way, it meets the default user expectation to just work on a Matrix, and at the same time is flexible to work on any Matrix if given the proper input of eigenvalues. Sound reasonable?

Beakerboy:
If we have multiple eigenvalue calculating algorithms, the eigenvector method will have to know which is the best to use. If the user determines the best for their situation, they can provide those values to this method.

markrogoyski:
Btw, I moved the Eigenvector method to its own class, because I expect the Eigenvalue class to grow as more methods are added.
Also, I added a eigenvalues and eigenvectors methods on the Matrix object which allow you to specify the eigenvalue method, and for now, default to polynomial roots if one is not specified (Not that it could default to anything else at this time :).

PR #234, opened 2017-06-06
Bitwise addition

Beakerboy:
This basically treats an int as if it were an unsigned int. The MSB is treated as a value bit instead of a signing bit.

### DIFF
--- a/src/Functions/Boolean.php
+++ b/src/Functions/Boolean.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace MathPHP\Functions;
+
+class Boolean
+{
+
+    /**
+     * Add two ints ignoring the signing bit.
+     *
+     * 8 bit examples:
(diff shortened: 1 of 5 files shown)


src/Functions/Boolean.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace MathPHP\Functions;
+
+class Boolean
```
markrogoyski: I'm not sure that Boolean is the right name for this class. I think of Boolean AND gates and OR gates etc. For these operations, I think the common term would be bitwise operations. So maybe Bitwise, or BitwiseOperation, or even BitwiseArithmetic. What do you think?

Beakerboy: Sounds good. I accidentally meged a development branch into this patch instead of vice-versa. I'll resubmit the patch in a minute.

PR #235, opened 2017-06-07
Bitwise Addition

Beakerboy:


tests/Functions/BitwiseTest.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public function dataProviderForBitwiseAdd()
+    {
+        return [
+            [
+                1, 1, [false, 2]
+            ],
+            [
+                1, -1, [true, 0]
```
markrogoyski: Why would overflow be true here?

tests/Functions/BitwiseTest.php
```diff
@@ -0,0 +1,46 @@
...
+            ],
+            [
+                \PHP_INT_MAX, 1, [false, \PHP_INT_MIN]
```
markrogoyski: Why is overflow false here?

tests/Functions/BitwiseTest.php
```diff
@@ -0,0 +1,46 @@
...
+            ],
+            [
+                -1, -1, [true, -2]
```
markrogoyski: Why is overflow true here?

src/Functions/Bitwise.php
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @return array
+     *    'overflow' is true if the result is larger than the bits in an int
+     *    'value' is the result of the addition ignoring any overflow.
+     */
+    public static function bitwiseAdd(int $a, int $b): array
+    {
+        if (is_int($a + $b)) {
+            $sum = $a + $b;
+            $overflow = (($a < 0 || $b < 0) && $sum >= 0) || ($a < 0 && $b < 0);
```
markrogoyski: I'm not sure I understand the definition of overflow here. If the result of $a + $b is an int, then by definition, there should be no integer overflow, correct?

Beakerboy: This is bit overflow, not integer overflow. The comment block for the method gives some bitwise examples. In the 4 bit world, 0b1111 + 0b0001 will overflow...however, in decimal, this in -1 + 1...which would not overflow. The bits overflow while the ints do not.

markrogoyski: OK. Thanks for the explanation. I get it now. I didn't see any binary numbers or bit shifting so I was a bit (pun intended) confused. It's interesting that you can figure out the overflow without converting to binary/2's complement and doing the operation just with this conditional statement.

Beakerboy: The textual description of this statement is...if you add a negative and positive number together, and the result is positive, the MSB on the negative number had to have been shifted off. Anytime two negative numbers are added, the MSB on both is a '1', and 1+1 = 10, so there is a bit overflow. It took me a while to make this statement it as compact as I could.

markrogoyski: Thanks again for the explanation. Do you mind adding a little more documentation and/or comments about the three cases to explain what is going on and why is doing what it is doing. Thanks.

markrogoyski:
Thanks for the pull request, and thanks for adding the explanatory documentation. I refactored a little bit to make the function a little cleaner, and changed the method name to just add as bitwiseAdd was redundant with the class name since it is a static method.

PR #236, opened 2017-06-09
BigInt v.000001 pre-alpha demo (Work in progress)

Beakerboy:
Here is a minimum implementation of a working 128-bit signed int. I am not proposing you merge this into develop, but you can at least look it over and ask questions and add comments. So far everything seems to work, although I need to add more unit tests since it's not at 100% coverage, and some of the methods have not been pushed to all edge cases.

Feel free to make changes and submit your own pull requests against my branch as well.

I have another branch with more methods in the works, like __toString(),  intdiv($c), and mod($c).

### DIFF
--- a/src/Number/BigInt.php
+++ b/src/Number/BigInt.php
@@ -171,7 +171,7 @@ class BigInt implements ObjectArithmetic
             $first = $first << 1;
             $second = $second << 1;
             if ($this->getBit(63 - $i) == 1) {
-                $second = Bitwise::bitwiseAdd($second, 1);
+                $second = Bitwise::add($second, 1);
             }
             if ($this->getBit(127 - $i) != $sign_bit) {
                 throw new Exception\OutOfBoundsException('Bitshift out of bounds');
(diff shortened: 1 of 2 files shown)


src/Number/BigInt.php
```diff
@@ -0,0 +1,345 @@
...
+     * Providing an array will create a BigInt with the bit squence of
+     * decbin($array[1]) . decbin($array[0])
+     *
+     * @param mixed $v
+     */
+    public function __construct($v)
+    {
+        $this->value = array_fill(0, self::WORDS, 0);
+        // Determine if this is 32 or 64 bit OS
+        $this->word_size = strlen(decbin(-1));
```
markrogoyski: Maybe something like this will make the intent clearer:
```php
$this->word_size = \PHP_INT_SIZE * 8;
```

Beakerboy: Good catch

Beakerboy:
Any general feedback on high level design and layout? I have intdiv(), mod(), pow(), and __toString() all working now too. __toString() is in base 10.

markrogoyski:
Sorry for the delay in providing feedback. I think I need to research this topic a little bit to become familiar with specific implementation strategies. Just curious, are you basing this design on some reference design?

One question I had was about the various constructor input types. Is it not a valid use case to have an integer string in base 10? For example:
```php
$int = new BigInt('9223372036854775807111122');
```

Once you push __toString, I'll write some unit tests and start to play around with the implementation.

Thanks.

src/Number/BigInt.php
```diff
@@ -0,0 +1,345 @@
...
+     *
+     */
+    public function add($c): BigInt
+    {
+        $type = gettype($c);
+        if ($type == 'integer') {
+            $c = new BigInt($c);
+        }
+        if ($c instanceof BigInt) {
+            $first = Bitwise::Add($this->value[0], $c->get(0));
```
markrogoyski: The public get method is not necessary. PHP has sort of an undocumented _family and friends_ object relationship where objects of the same type can access other objects' private attributes. So no need to expose the implementation detail of how the BigInt is represented and just manually access $c's value property directly within a BigInt method.

Beakerboy:
A base 10 constructor is much more difficult than a binary, hex, or octal constructor. Now that I have integer division and modular arithmetic working, I can do a base 10 constructor.

The get() method is kind of a hack, and I assumed there was a better way of doing this. I'll have to look at ways of changing it.

PR #237, opened 2017-06-13
Update setup

leohajder:
Add composer require command for Symfony projects installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,12 @@ Composer will install MathPHP inside your vendor folder. Then you can add the fo
 require_once(__DIR__ . '/vendor/autoload.php');
 ```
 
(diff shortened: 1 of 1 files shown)


markrogoyski:
Hi,
Thanks for your interest in MathPHP! And thanks for the pull request.

I took this pull request to mean that the README should have the command-line composer require command in addition to manually editing a composer.json file. I went ahead and added a similar line in the README's setup, but I worded it differently, as it isn't necessarily Symfony specific. So I won't merge this pull request.

I merged the change to master so you can see it on the main page.

Thanks for your suggestion and thanks for using MathPHP!

leohajder:
Cool, I'm glad to help. Maybe remove the bash highlighting from the command because the * looks strange.

markrogoyski:
Updated. Thanks.

PR #240, opened 2017-07-19
Add Kernel Density Estimation

Beakerboy:


src/Statistics/KernelDensityEstimation.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * KernelDensity
+     *
+     * @param float $x the value to evaluate
+     *
+     * @return float the kernel density estimate at $x
+     */
+    public function KernelDensity($x)
```
markrogoyski: Would ```evaluate``` or ```estimate``` be a better name for this method?

markrogoyski:
Thanks for the pull request and implementing the kernel density estimation.

Do you think it would make sense to provide some standard kernel function options for the user to select, in addition to allowing a custom callable function and the default of standard normal?

What did you use to create the unit test data?
Thanks.

Beakerboy:
Unit tests were calculated in excel. Octave has not implemented the matlab ksdensity() or fitdist() functions. I generated the normal data in Excel, and ran it through a similar algorithm.

````
    public static function getKernelFunction(string $kernel_name)
    {
        switch ($kernel_name) {
            case 'normal':
              $kernel = ['StandardNormal', 'pdf'];
              break;
            case 'tricube':
                $kernel = function($x) {
                    ....
                };
                break;
            case 'Epanechnikov':
                $kernel = function($x) {
                   ....
                };
                break;
        }
        return $kernel;
    }
````
And then then could we use it like:

````
    $KDE = new KernelDensityEstimation($data, null, KernelDensityEstimation::getKernelFunction('tricube');
````

File: src/Statistics/KernelDensityEstimation.php (hunk 1 above)
Comment (Beakerboy) [inline src/Statistics/KernelDensityEstimation.php]: sounds good.

**Comment by markrogoyski** [comment]
Constructing the KD estimator is based on the initial data. Then evaluating is based on running the kernel function. I can imagine a use case where someone wants to evaluate against the same data using different kernel functions (is this common?).

So, it might make sense to separate construction of the KDE object and selecting a kernel function.
What do you think about this example:
```php
 // Defaults to standard normal
$kde = new KernelDensityEstimation($data, $h);
for ($xs as $x) {
   $estimate = $kde->evaluate($x);
   ...
}

// Provided standard functions
$kde->setKernelFunction(KernelDensityEstimation::EPANECHNIKOV);
for ($xs as $x) {
   $estimate = $kde->evaluate($x);
   ...
}

// Custom functions
$kde->setKernelFunction(function ($x) { ... });
for ($xs as $x) {
   $estimate = $kde->evaluate($x);
   ...
}

```

Beakerboy:
I can see an advantage to having setters for $h and $kernel. That way they can be changed without creating a new object. 

````
    public function setH(float $h)
    {
        // Check bounds of $h
        $this->h = $h;
    }

    const NORMAL = 0;
    const EPANECHNIKOV = 1;
    const TRICUBE = 2;

    /**
     * Set The Kernel Function
     *
     * If the parameter is a string, check that there is a function with that name 
     * in the "library". If it's a callable, use that function.
     */
    public function setKernelFunction($kernel)
    {
        if (is_int($kernel)) {
            switch ($kernel) {
                case self::EPANECHNIKOV:
                    $kernel = function($x) {
                       ....
                    };
                    break;
                case self::TRICUBE:
                    $kernel = function($x) {
                       ....
                    };
                    break;
                default:
                    $kernel = ['StandardNormal', 'pdf'];
                break;
            }
            $this->kernel = $kernel;
        } else if (is_callable($kernel)) {
            $this->kernel = $kernel;
        } else {
            // throw an exception
        }
    }

**Comment by markrogoyski** [comment]
I'd probably rename ```setH()``` to ```setBandwidth``` or something. Other than that your approach looks good. Do you want to add in that functionality with tests and then we'll merge that for an initial version?

**Comment by markrogoyski** [comment]
I'm trying to write some unit tests for the KDE, and want to create test data expected results using a third-party source, but I'm having trouble getting results to match against a trusted third-party source. I tried R and Python's SciPy, but can't quite get the results to match. I looked over the math in the ```evaluate``` function, and it looks correct. And we're pretty sure the StandardNormal PDF works I think.

Can you look over what I'm doing in Python vs. PHP and see if there is something obvious I'm doing wrong?

Python SciPy
```python
>>> from scipy import stats
>>> points
(-2.76, -2.59, -2.57, -2.4, -1.9, -1.77, -1.33, -1.27, -1.26, -1.19, -1.13, -1.1, -1.09, -1.05, -1.03, -0.9, -0.87, -0.87, -0.85, -0.83, -0.78, -0.71, -0.62, -0.61, -0.6, -0.57, -0.55, -0.51, -0.5, -0.5, -0.47, -0.46, -0.46, -0.45, -0.44, -0.32, -0.28, -0.27, -0.25, -0.23, -0.21, -0.21, -0.2, -0.16, -0.15, -0.11, -0.094, -0.07, -0.06, -0.04, 0, 0.03, 0.04, 0.06, 0.06, 0.09, 0.1, 0.12, 0.15, 0.15, 0.18, 0.22, 0.22, 0.23, 0.24, 0.3, 0.31, 0.34, 0.34, 0.38, 0.38, 0.44, 0.52, 0.53, 0.56, 0.57, 0.61, 0.61, 0.69, 0.7, 0.75, 0.79, 0.79, 0.79, 0.8, 0.83, 0.85, 0.99, 1.08, 1.23, 1.23, 1.23, 1.25, 1.29, 1.32, 1.34, 1.43, 1.49, 1.62, 1.75)
>>> kde = stats.gaussian_kde(points, 0.3932)
>>>
>>> kde(-3.939)
array([  8.60829883e-05])
>>> kde(-2.222)
array([ 0.04138222])
>>> kde(-0.505)
array([ 0.36562621])
>>> kde(0)
array([ 0.43200043])
>>> kde(1.212)
array([ 0.19113825])
>>> kde(2.929)
array([  9.09492454e-05])
```
Unit test
```php
    public function testAgainstSciPy()
    {
        $KDE = new KernelDensityEstimation($this->data);
        $h   = 0.3932;
        $KDE->setBandwidth($h);
        $KDE->setKernelFunction(KernelDensityEstimation::NORMAL);

        $x = -3.939;
        $this->assertEquals(8.60829883e-05, $KDE->evaluate($x), '', 0.00001);

        $x = -2.222;
        $this->assertEquals(0.04138222, $KDE->evaluate($x), '', 0.00001);

        $x = -0.505;
        $this->assertEquals(0.36562621, $KDE->evaluate($x), '', 0.00001);

        $x = 0;
        $this->assertEquals(0.43200043, $KDE->evaluate($x), '', 0.00001);

        $x = 1.212;
        $this->assertEquals(0.19113825, $KDE->evaluate($x), '', 0.00001);

        $x = 2.929;
        $this->assertEquals(9.09492454e-05, $KDE->evaluate($x), '', 0.00001);
    }
```

Unit test results
```
Evaluate x = -3.939
Failed asserting that 0.000169895343073927 matches expected 8.6082988300000004E-5.

Evaluate x = -2.222
Failed asserting that 0.042122201381434098 matches expected 0.041382219999999997.

Evaluate x = -0.505
Failed asserting that 0.36309951007398344 matches expected 0.36562621000000001.

Evaluate x = 0
Failed asserting that 0.42555701694570436 matches expected 0.43200042999999999.

Evaluate x = 1.212
Failed asserting that 0.1910833731679274 matches expected 0.19113825000000001.

Evaluate x = 2.929
Failed asserting that 0.00018361554921641045 matches expected 9.0949245400000003E-5.
```

Also, for reference, here is what R produces using density:
```
Call:
	density.default(x = points, bw = 0.39315925657134, kernel = "gaussian")

Data: points (100 obs.);	Bandwidth 'bw' = 0.3932

       x                y            
 Min.   :-3.939   Min.   :0.0001715  
 1st Qu.:-2.222   1st Qu.:0.0222563  
 Median :-0.505   Median :0.0730575  
 Mean   :-0.505   Mean   :0.1454348  
 3rd Qu.: 1.212   3rd Qu.:0.2689829  
 Max.   : 2.929   Max.   :0.4259432  
```

Thanks for your help.

Beakerboy:
I've never used python before, so this is new...It looks like this is not using the same kernel function because the standard deviation of the data is used in the calculation. I think if you change the kernel from Standard normal to normal, and use the standard deviation of the dataset, you will get the same numbers.
Something like this:
````
    $stdev = Descriptive::standardDeviation($data);
    $kernel =function($x) using $stdev ( return Normal::pdf($x, 0,$stdev);};
````

markrogoyski:
I think you are right! I changed to use Normal with the standard deviation and all my tests pass now. Thanks.

Do you think we should add a Normal built-in option in addition to standard normal in the ```setKernelFunction``` method? And maybe change the constants to be ```STANDARD_NORMAL``` and ```NORMAL``` respectively to give clearer intent of what is going on? Also, should the default kernel function be StandardNormal or normal?

Thanks again for your help!

Beakerboy:
Looking at the way the code is organised, it's more that their definition of bandwidth is the factor to multiply with the standard deviation...at least when it's given a scaler. 

I'll leave those choices up to you dear leader ;-)
My vote is to leave it alone since our strategy matches the Wikipedia article. Their's may match some other source.

Beakerboy:
Since their function works for multidimensional data as well, it's possible that the math is simpler for higher dimensions using their strategy.

markrogoyski:
If the dataset has a standard deviation that is not close to zero, will using the StandardNormal as the default kernel function give a reasonable result?

markrogoyski:
How about we leave the default as StandardNormal and add a convenience constant option for Normal similar to EPANECHNIKOV etc.?

Beakerboy:
If a user has data from scipy that they want to move over to this library, they have two options, change the bandwidth, or change the Kernel. I propose that we keep our library the same and just have users change the bandwidth. To make it even simpler, we should also add the bandwidth estimation function that scipy defaults to, since most users likely are not specifying the bandwidth. We should also be more clear on how the bandwidth is being estimated so people know if it is or is not including the standard deviation.

markrogoyski:
My only interest in the SciPy function was to make unit test data to test against. I'm not interested in copying other libraries (learning from them and getting ideas is always a good idea though). Let's do what we think is right for this library. 

When I write unit tests, I always try to use a trusted third party. Maybe this is examples from a textbook or online source, or an R function or online calculator, etc. For the case of kernel density estimation, I couldn't really find any good full worked out examples, so the next best thing was a different library. I avoid reimplementing the implemented code in Excel or something because if a mistake was made, and the same mistake is repeated in Excel, the test data isn't exactly testing that your code is correct.

Beakerboy:
That's all good stuff. I usually use octave-online as a double check, but their function is not available.

PR #244, opened 2017-08-22
RationalNumber Object

Beakerboy:
A rational number can be expressed as a fraction. Using the rational number object allows a user to express non-integer values with exact precision, and perform arithmetic without floating point errors.

### DIFF
--- a/src/Number/RationalNumber.php
+++ b/src/Number/RationalNumber.php
@@ -1,0 +1,299 @@
+<?php
+
+namespace MathPHP\Number;
+
+use MathPHP\Algebra;
+use MathPHP\Exception;
+use MathPHP\Functions\Special;
+
+/**
+ * Rational Numbers
+ *
+ * https://en.wikipedia.org/wiki/Rational_number
+ * A rational number can be expressed as a fraction. Using the rational number object allows a user to
+ * express non-integer values with exact precision, and perform arithmetic without floating point
+ * errors.
+ */
+class RationalNumber implements ObjectArithmetic
+{
+    /**
+     * Whole part of the number
+     * @var int
+     */
+    protected $whole;
+    
+    /**
+     * Numerator part of the fractional part
+     * @var int
+     */
+    protected $numerator;
+    
+    /**
+     * Denominator part of the fractional part
+     * @var number
+     */
+    protected $denominator;
+    
+    /**
+     * Constructor
+     *
+     * to do: How to handle negative numbers in various positions.
+     * @param int $w whole part
+     * @param int $n numerator part
+     * @param int $d
+     */
+    public function __construct(int $w, int $n, int $d)
+    {
+        list($w, $n, $d) = self::normalize($w, $n, $d);
+        $this->whole = $w;
+        $this->numerator = $n;
+        $this->denominator = $d;
+    }
+    
+    /**
+     * String representation of a rational number
+     * 5 6/7, 456079/13745859, etc.
+     *
+     * @return string
+     */
+    public function __toString(): string
+    {
+        $sign = "";
+        $whole = "";
+        $fraction = "";
+        if (Special::sgn($this->whole) == -1 || Special::sgn($this->numerator) == -1) {
+            $sign = "-";
+        }
+        if ($this->whole !== 0) {
+            $whole = abs($this->whole);
+        }
+        if ($this->numerator !== 0) {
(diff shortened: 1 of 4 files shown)


src/Number/RationalNumber.php
```diff
@@ -0,0 +1,299 @@
...
+            $fraction = $this->toSuperscript(abs($this->numerator)) . "/" . $this->toSubscript($this->denominator);
+        }
+        $string = $sign . $whole . $fraction;
+        if ($string == "") {
+            $string = "0";
+        }
+        return $string;
+    }
+    
+    private static function toSuperscript(int $i)
```
markrogoyski: Why are these three super/subscript functions static? Make them object methods and change ```self::``` invocations to ```$this->```.

src/Number/RationalNumber.php
```diff
@@ -0,0 +1,299 @@
...
+
+/**
+ * Rational Numbers
+ *
+ * https://en.wikipedia.org/wiki/Rational_number
+ * A rational number can be expressed as a fraction. Using the rational number object allows a user to
+ * express non-integer values with exact precision, and perform arithmetic without floating point
+ * errors.
+ */
+class RationalNumber implements ObjectArithmetic
```
markrogoyski: How about naming it ```Rational``` since Number is in the namespace. That will also follow the convention of ```Complex``` and the not-yet-merged ```BigInt```.

src/Number/RationalNumber.php
```diff
@@ -0,0 +1,299 @@
...
+        $working_value = $i;
+        for ($j = $number_of_chars - 1; $j >= 0; $j--) {
+            $int = intdiv($working_value, 10 ** $j);
+            $return_string .= $chars[$int];
+            $working_value -= $int * 10 ** $j;
+        }
+        return $return_string;
+    }
+    
+    public function toFloat()
```
markrogoyski: I was thinking about method naming conventions. What would be more natural, ```toFloat``` or ```asFloat```? I  like ```asFloat``` because it tends to have a more natural expressiveness. For example:
```php
print($number->asFloat());
```
vs.
```php
print($number->toFloat());
```
However, the built-in ```__toString``` is doing basically the same kind of operation and is the established naming convention for this. So I'm OK with ```toFloat``` for consistency.

src/Number/RationalNumber.php
```diff
@@ -0,0 +1,299 @@
...
+    /**
+     * Addition
+     *
+     * @param mixed $c
+     *
+     * @return RationalNumber
+     *
+     * @throws Exception\IncorrectTypeException if the argument is not numeric or Rational.
+     */
+    public function add($c): RationalNumber
```
markrogoyski: Why ```$c``` as the parameter name in add and multiply? Subtract and divide both use ```$r```.

markrogoyski:
Thanks for the pull request. This is great stuff.

We could use this to make a ```NumericallyStableMatrix``` class that uses Rational numbers for all the elements.

tests/Number/RationalNumberTest.php
```diff
@@ -0,0 +1,328 @@
...
+     * @dataProvider dataProviderForToString
+     * @param        number $w
+     * @param        number $n
+     * @param        number $d
+     * @param        string $string
+     */
+    public function testToString($w, $n, $d, string $string)
+    {
+        $number = new RationalNumber($w, $n, $d);
+        $this->assertEquals($string, $number->__toString());
```
markrogoyski: Use ```assertSame``` in this case because you are testing value and type.

tests/Number/RationalNumberTest.php
```diff
@@ -0,0 +1,328 @@
+<?php
+namespace MathPHP\Tests\Number;
+
+use MathPHP\Number\RationalNumber;
+use MathPHP\Exception;
+
+class RationalNumberTest extends \PHPUnit_Framework_TestCase
```
markrogoyski: Nice unit tests.

src/Number/RationalNumber.php
Beakerboy: I just made them static because they do not depend on the state of the object. I'll change them.

Beakerboy:
Because this class implements ObjectArithmetic, the ObjectMatrix would already be able to use this as its data type and provide the numerical stability you want.

PR #248, opened 2017-09-28
Transitioning distributions to non-static

Beakerboy:
Here is a first step. I added a generic parent-level constructor, and transitioned the Beta distribution to be non-static.

### DIFF
--- a/src/Probability/Distribution/Continuous/Beta.php
+++ b/src/Probability/Distribution/Continuous/Beta.php
@@ -11,14 +11,21 @@ use MathPHP\Functions\Support;
 class Beta extends Continuous
 {
     /**
-     * Distribution parameter bounds limits
+     * Distribution support bounds limits
      * x ∈ [0,1]
+     * @var array
+     */
+    const SUPPORT_LIMITS = [
+        'x' => '[0,1]',
+    ];
+
+    /**
+     * Distribution parameter bounds limits
      * α ∈ (0,∞)
      * β ∈ (0,∞)
      * @var array
      */
-    const LIMITS = [
-        'x' => '[0,1]',
+    const PARAMETER_LIMITS = [
         'α' => '(0,∞)',
         'β' => '(0,∞)',
     ];
@@ -53,8 +60,7 @@ class Beta extends Continuous
      */
     public function pdf($x)
     {
-        $limit['x'] = static::LIMITS['x'];
-        Support::checkLimits($limit, ['x' => $x]);
+        Support::checkLimits(self::SUPPORT_LIMITS, ['x' => $x]);
 
         $α = $this->α;
         $β = $this->β;
@@ -75,8 +81,7 @@ class Beta extends Continuous
      */
     public function cdf($x)
     {
-        $limit['x'] = static::LIMITS['x'];
-        Support::checkLimits($limit, ['x' => $x]);
+        Support::checkLimits(self::SUPPORT_LIMITS, ['x' => $x]);
 
         $α = $this->α;
         $β = $this->β;
--- a/src/Probability/Distribution/Continuous/Beta.php
+++ b/src/Probability/Distribution/Continuous/Beta.php
@@ -28,7 +28,18 @@ class Beta extends Continuous
 
     /** @var number Shape Parameter */
     protected $β;
-    
+
+    /**
+     * Constructor
+     *
+     * @param number $α shape parameter α > 0
+     * @param number $β shape parameter β > 0
+     */
+    public function __construct($α, $β)
+    {
+        parent::_construct($α, $β);
+    }
+
     /**
      * Probability density function
      *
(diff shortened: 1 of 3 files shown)


src/Probability/Distribution/Continuous/Continuous.php
```diff
@@ -1,8 +1,35 @@
...
+use MathPHP\Functions\Support;
+use MathPHP\Exception;
+
 abstract class Continuous extends \MathPHP\Probability\Distribution\Distribution
 {
+    const LIMITS = [];
+
+    protected $params;
+    
+    public function __construct(...$params)
```
markrogoyski: I'm not convinced this is better than each distribution having its own constructor. With a unique constructor for each distribution, you will have a clearly defined method signature, and you have the ability to use type declarations for better type safety.

If each constructor set its own object attributes, then you can directly use them in the cdf and pdf methods. Otherwise, each method has additional preparation work to do: (```$α = $this->params["α"];```) for instance. You save code on the individual constructors, at the trade off of adding additional code into all the computation methods. I think I'd prefer any setup to happen in the constructor and let's try to have the computation methods as clean as possible.

I think it makes sense to move the checking of limits into a shared common class. Each individual constructor could call ```parent::__construct``` and share that code.

What do you think?

Beakerboy: Are there any distributions that use type declarations? 

Adding ```$this->$key = $params[$i];``` to the foreach loop in ```Continuous::__construct()``` would save each parameter to its own object attribute, and get rid of ```$this->params```. In general, we both often do something like ```$α = $this->α;``` on the top of many functions anyways, so I didn't think there is any extra prep-work. If this parent constructor is not good for some specific distribution, if can always be overridden.

markrogoyski: Yeah, you are right that often I like to work with a shorter local variable than a ```$this->something``` for conciseness and do just that.

My main issue is with not having a method signature. The method signature clearly shows you what the object's parameters without having to look at external documentation. If you use an IDE like PHPStorm, having a well-defined method signature brings tons of benefits.

Could we retain the benefit of a shared parent constructor doing the common work and still have individual method signatures?

src/Probability/Distribution/Continuous/Beta.php
```diff
@@ -76,10 +74,9 @@ public static function cdf($x, $α, $β)
      *
      * @return number
      */
-    public static function mean($α, $β)
+    public function mean()
     {
-        Support::checkLimits(self::LIMITS, ['α' => $α, 'β' => $β]);
 
-        return $α / ($α + $β);
+        return $this->α / ($this->α + $this->β);
```
markrogoyski: You are right. I definitely prefer this:
```return $α / ($α + $β);```

over
```return $this->α / ($this->α + $this->β);```

File: src/Probability/Distribution/Continuous/Continuous.php (hunk 1 above)
Comment (Beakerboy) [reply src/Probability/Distribution/Continuous/Continuous.php]: The only way I can think to add individual method signatures would be to add a dummy constructor:
```
class Beta
{
    public function __construct($α, $β)
    {
        parent::_construct($α, $β);
    }
}
````

Comment (markrogoyski) [reply src/Probability/Distribution/Continuous/Continuous.php]: I think the constructor that just calls its parent is better than none. You get the method signature, you can add a docBlock which documentation, you have the option for a type declarations, and you can see how to use the function when you look at the code. Without the constructor, it is not obvious how you construct an object. You can also add the object parameter declarations with dockBlocks as well.

Even if all the work happens in the parent, it is nice to have a constructor that shows how to use it.

File: src/Probability/Distribution/Continuous/Beta.php [hunk 3]
```diff
@@ -22,24 +22,42 @@ class Beta extends Continuous
...
-     * @param number $β shape parameter β > 0
      * @param number $x x ∈ (0,1)
      *
      * @return float
      */
-    public static function pdf($x, $α, $β)
+    public function pdf($x)
     {
-        Support::checkLimits(self::LIMITS, ['x' => $x, 'α' => $α, 'β' => $β]);
+        $limit['x'] = static::LIMITS['x'];
```
markrogoyski: I think we could modify ```Support::checkLimits``` slightly so this extra step is not necessary. What do you think about modifying the ```Support::checkLimits``` function to only check that every parameter being checked has a limit defined for it, rather than the current check which is that every parameter has a limit defined AND every limit has a parameter being checked.

Beakerboy: Alternatively, we could have separate arrays, one for support and one for parameters.

markrogoyski: That works too. Probably a better idea since we keep the check that every limit needs to have a parameter passed in. Thanks.

markrogoyski:
Thanks for the pull request and getting this started.

PR #249, opened 2017-10-03
Distribution update (#23)

Beakerboy:
Not all...but most.


markrogoyski:
Thanks for the pull request and leading this refactor effort.

PR #252, opened 2017-10-03
Beginning Discrete Distributions

Beakerboy:
Moved the default constructor up one level, and refactored Binomial


markrogoyski:
Thanks for starting on the discrete distributions.

Issue #247, opened 2017-09-12
Refactor distributions to objects


balping:
Currently distributions are static interfaces that follow an approach borrowed by functional languages. Although you can use PHP in a functional style, it is clearly object oriented and in the case of distributions, we could use the benefits of OOP.

An instance of a distribution could be a distribution function with  given parameters. And by 'function' I mean the whole abstract pdf/cdf functions, without evaluated for a certain value. Probably the easiest way to explain this is via an example. So take this example of the binomial distribution:

```php
// current
$v1 = Binomial::pmf($n, $r, $P);
$v2 = Binomial::cdf($n, $r, $P);
```

would be changed to this:

```php
// planned
$binom = new Binomial($r, $P);
// we defined a distribution object with given parameters, now we can calculate things
$v1 = $binom->pmf($n);
$v2 = $binom->cdf($n);
```

This approach  would have the following benefits:
* separating independent variables from parameters
* a consistent api across different distributions, maybe forced by an interface
* consistency with categorical distribution that works in an object oriented way already
* possibility to validate parameters at initialisation and not at computation (eg. now in recursive cdf functions the same parameters are validated many times unnecessarily)
* possibility to add any number of helper methods (eg. min/max of independent variable, see below)
* it would make it _much_ easier and more efficient to implement random number generation on discrete distributions:
  -  min/max values of the independent variable (_0_ and _r_ for binomial) 
 could be queried via a simple method call
  - lookup tables could be cached if we need multiple random values
  - the api could be as simple as this: `$rand = $binom->random();`

cons:
* heavily breaking change, but this library is pre 1.0, so I wouldn't consider this as a serious issue

I'm absolutely willing to do this refactoring (and the implementation of random on discrete distributions)  but I wouldn't  start working on it without the approval of the maintainers, because I wouldn't like to waste my time.

markrogoyski:
Hi @balping 
Thank you for your interest in MathPHP and making the library better.

This has been suggested before, and I think it makes sense.

Generally, if something is deterministic, depending only on its input, then we usually make it static. For example:
```php
Algebra::gcd(8, 12)
```

If something has state, or if multiple things can be computed from some initial values, then it is an object with methods. For example, Matrix, Vector, Regressions, etc.

In the case of probability distributions, the original thinking was that the PDF or CDF are deterministic based on all the input parameters. However, you could think of it as a distribution is made up of parameters, and then using those parameters, you can compute the PDF or CDF using different support values.

[...]

PR #255, opened 2017-10-12
Distributions: continuous interface

balping:
https://github.com/markrogoyski/math-php/issues/247

* Added a continuous interface
* unified cdf and pdf signatures to have float type
* removed some unnecessary docblock lines

Please verify that it makes sense to say that the dirac delta distribution has a mean of 0.
https://github.com/balping/math-php/blob/2bbb17dd8f19cd0f38641696efa479103f16a8b5/src/Probability/Distribution/Continuous/DiracDelta.php#L98

### DIFF
--- a/src/Probability/Distribution/Continuous/Beta.php
+++ b/src/Probability/Distribution/Continuous/Beta.php
@@ -54,11 +54,11 @@ class Beta extends Continuous
      * pdf = --------------
      *           B(α,β)
      *
-     * @param number $x x ∈ (0,1)
+     * @param float $x x ∈ (0,1)
      *
      * @return float
      */
-    public function pdf($x)
+    public function pdf(float $x)
     {
         Support::checkLimits(self::SUPPORT_LIMITS, ['x' => $x]);
 
@@ -76,11 +76,11 @@ class Beta extends Continuous
      *
      * cdf = Iₓ(α,β)
      *
-     * @param number $x x ∈ (0,1)
+     * @param float $x x ∈ (0,1)
      *
      * @return float
      */
-    public function cdf($x)
+    public function cdf(float $x)
     {
         Support::checkLimits(self::SUPPORT_LIMITS, ['x' => $x]);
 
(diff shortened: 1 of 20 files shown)


Beakerboy:
I have not done the math, but my initial guess is the mean is undefined.

markrogoyski:
Hi,
Thank you for the pull request!

What is the advantage of the Continuous interface? Every continuous distribution extends MathPHP\Probability\Distribution\Continuous so you could use a type declaration of Distribution\Continuous already. Also, since not every distribution has or implements a mean, not every continuous distribution can implement the Continuous interface.

For the dirac delta distribution mean, if you have a definitive source that defines the mean, please put that in the documentation. And add unit tests.

Thanks.

Beakerboy:
After more thought...maybe the mean is 0. If you take random draws from the distribution, their average will always be 0. I think proving it through integration is tricky (if even possible)

balping:
The inverse function defined in the abstract class, depends on cdf, pdf and mean function. The interface enforces that they are defined, so calling inverse (and rand) doesn't lead to unexpected function not defined errors. The interface also enforces consistency in api across distributions.

I asked a question about the mean of Dirac delta on [math stackoverflow](https://math.stackexchange.com/questions/2470914/does-the-dirac-delta-distribution-have-a-mean/2470921#2470921).  It seems that the mean is indeed 0, so I'll add a test to get that 100% coverage back.

markrogoyski:
Let's note use the word ```Interface``` in the interface and use something more descriptive. Since the interface doesn't describe all Continuous distributions, would it make more sense to call it ```InvertibleDistribution``` or something like that?

balping:
Up until now, it does describe all continuous distributions.

I don't like the `Interface` suffix either, normally I would have gone for `Continuous`, but that's already taken

markrogoyski:
Thanks for the pull request. I renamed the interface ```ContinuousDistribution```.

PR #274, opened 2018-11-09
Added Grubbs Test for Outliers

Beakerboy:
* Added a function to calculate the Grubbs statistic from a set of data
 * Added a function to calculate the critical Grubbs value given a confidence and number of data points
 * Added a test case using the NIST data for this function

### DIFF
--- a/src/Statistics/Significance.php
+++ b/src/Statistics/Significance.php
@@ -550,6 +550,6 @@ class Significance
     {
         $studentT = new StudentT($n - 2);
         $T = $studentT->inverse($𝛼 / $n);
-        return ($n - 1) * $T / sqrt($n * ($n - 2 + $T ** 2))
+        return ($n - 1) * $T / sqrt($n * ($n - 2 + $T ** 2));
     }
 }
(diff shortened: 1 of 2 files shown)


markrogoyski:
Thanks for the pull request! Do we think this belongs in ```Significance```, or might it be better organized in a ```Statistics\Outliers``` namespace or something like that?

Would it be useful to add a mechanism to "clean" a dataset of outliers using the Grubbs' test? Wikipedia describes it like this:

> Grubbs' test detects one outlier at a time. This outlier is expunged from the dataset and the test is iterated until no outliers are detected.

Also, should there be a check on the size of the dataset? Also from Wikipedia:

> the test should not be used for sample sizes of six or fewer since it frequently tags most of the points as outliers.

How do we think people would use this? Is it better to give lots of options or make the interface very specific towards one way of using it? Would it be useful to have a boolean method that simply response yes or no if it has outliers according to the test? In other words, have the library answer the question:

* H0: There are no outliers in the data set
* Ha: There is exactly one outlier in the data set

Beakerboy:
I don't know how people would use it. I use it for interlab testing to see if one analyst's results are significantly different than the bulk when performing the same test on the the same sample. I just flag if an outlier exists or not.

It would probably be good if it behaves like a T-Test, and give a p-value. This would require taking the inverse of the CriticalGrubbs function.

Beakerboy:
maybe it's a little philosophical, but there seem to be overlaps in "significance" and "outlier". An outlier is a subset of data that does not match the rest. A t-test checks two sets of data to see how likely they are to be a part of the same set. I'm fine with having a separate namespace, especcially given that there are many different univariate outlier algorithms that can be implemented.

PR #277, opened 2018-11-16
Outliers

Beakerboy:
Created two outlier tests in a new namespace

### DIFF
--- a/src/Statistics/Outlier.php
+++ b/src/Statistics/Outlier.php
@@ -1,0 +1,140 @@
+<?php
+namespace MathPHP\Statistics;
+
+use MathPHP\Functions\Map\Single;
+use MathPHP\Probability\Distribution\Continuous\StandardNormal;
+use MathPHP\Probability\Distribution\Continuous\StudentT;
+use MathPHP\Statistics\Average;
+use MathPHP\Statistics\Descriptive;
+
+/**
+ * Tests for outliers in data
+ *  - Grubbs Test
+ *  - Tietjen-Moore Test
+ */
+class Outlier
+{
+    /**
+     * The Grubbs Statistic (G) of a series of data
+     *
+     * G is the largest z-score for a set of data
+     * The statistic can be calculated, looking at only the maximum value ("upper")
+     * the minimum value ("lower"), or the data point with the largest residual ("two")
+     *
+     * @param array $data
+     * @param $tails ("upper" "lower", or "two")
+     *
+     * @return float
+     */
+    public static function GrubbsStatistic(array $data, string $tails = "two"): float
+    {
+        if ($tails != "upper" && $tails != "lower" && $tails != "two") {
+            //throw exception;
+        }
+        $μ = Average::mean($data);
+        $σ = Descriptive::standardDeviation($data);
+        
+        $difference = Single::multiply(Single::subtract($data, $μ), ($tails == "upper" ? 1 : -1));
+        $max = $tails == "two" ? max(Single::abs($difference)) : max($difference);
+        return $max / $σ;
+    }
+    
+    /**
+     * The critical Grubbs Value
+     * https://en.wikipedia.org/wiki/Grubbs%27_test_for_outliers
+     * https://www.itl.nist.gov/div898/handbook/eda/section3/eda35h1.htm
+     *
+     * The Critical Gubbs value is used to determine if a value in a set of data is
+     * likely to be an outlier.
+     *
+     * @param float $𝛼 Significance Level
+     * @param int $n Size of the data set
+     * @param int $tails (1 or 2) one or two-tailed test
+     *
+     * @return float
+     */
+    public static function CriticalGrubbs(float $𝛼, int $n, int $tails = 2): float
+    {
+        if ($tails < 1 || $tails > 2) {
+            //throw Exception;
+        }
+        $studentT = new StudentT($n - 2);
+        $T = $studentT->inverse($𝛼 / $n / $tails);
+        return ($n - 1) * sqrt($T ** 2 / $n / ($n - 2 + $T ** 2));
+    }
+
+    /**
+     * The Tietjen-Moore Statistic of a set of data
+     *
+     * https://www.itl.nist.gov/div898/handbook/eda/section3/eda35h2.htm
+     *
+     * @param array $data
+     * @param int $k The number of outliers to test for
+     * @param $tails ("upper" "lower", or "two")
+     *
+     * @return float
+     */
+    public static function TietjenMooreStatistic(array $data, int $k, string $tails = "lower"): float
+    {
+        $ybar = Average::mean($data);
+        $den = array_sum(Single::square(Single::subtract($data, $ybar)));
+        $n = count($data);
+        $num = 0;
+        $smaller_set = [];
+        if ($tails == "two") {
+            $z = Single::abs(Single::subtract($data, $ybar));
+            $kthlargest = Average::kthsmallest($z, $n - $k - 1);
+            foreach ($data as $value) {
+                if (abs($value - $ybar) <= $kthlargest) {
+                    $smaller_set[] = $value;
+                }
+            }
+        } elseif ($tails == "upper") {
+            $kthlargest = Average::kthsmallest($data, $n - $k - 1);
+            foreach ($data as $value) {
+                if ($value <= $kthlargest) {
+                    $smaller_set[] = $value;
+                }
+            }
+        } elseif ($tails == "lower") {
+            $kthlargest = Average::kthsmallest($data, $k);
+            foreach ($data as $value) {
+                if ($value >= $kthlargest) {
+                    $smaller_set[] = $value;
+                }
+            }
+        }
+        $ykbar = Average::mean($smaller_set);
+        $num = array_sum(Single::square(Single::subtract($smaller_set, $ykbar)));
+        return $num / $den;
+    }
+
+    /**
+     * The critical Tietjen-Moore Value
+     *
+     * The Critical Tietjen-Moore value is used to determine if multiple values in a set of data are
+     * likely to be outliers.
+     *
+     * The critical value is determined by simulation.
+     *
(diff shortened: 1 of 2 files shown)


Beakerboy:
If you rerun the failed build, it will work. One of the new methods estimated a parameter by simulation. I ran some offline simulations and the given value plus the error bandwidth should cover 95%+ of cases, but it will fail once in a while. We can either increase the bandwidth, or increase the number of simulations.

markrogoyski:
Thanks for working on this!

Now, a unit test should be deterministic. And a failure should be meaningful.

I don't know enough about the critical Tietjen-Moore calculation, but if it uses random values and a test fails because of that, wouldn't it also mean that it will give a wrong answer to someone using the function at random times?

Basically, we only want pure deterministic functions. In other words, functions that always return the same result any time they are called with a specific set of input values. If I can call it twice with the same input and get two different answers, I don't know if I trust what it does.

If this is just how the algorithm works and it is understood that work that way, that's fine I suppose, but we have to make the test resilient. A test should never fail for any reason other than the function is broken. So, think about how to make the unit test pass 100% of the time with the same input when the function works as expected.

Thanks.

Beakerboy:
NIST says the parameter is to be determined by simulation using 10000 simulations. Their example then goes on to use the value .317. I did extensive simulation of millions of random normal data and found the true value to be .315, so that’s what I used for the unit test. The funny thing is, in all my coding and troubleshooting it never failed. Of course the only time is does is when submitting it. 

I ran a million simulations and looked at the distribution of the estimated Ek value from each 10,000 simulation chunk, and picked the dead and of .005 to cover 99% of the area. Increasing the number of simulations to 100000 should bring the gap down to .0015, and a million simulations to .0005.

Beakerboy:
I removed Tietjen-Moore. As reference, here's how R handles some outlier tests. Maybe we could use this as guidance on how to specify input parameters, function names, or expected output.
(http://cran.cnr.berkeley.edu/web/packages/outliers/outliers.pdf)

markrogoyski:
I was able to use R to reproduce the value for the Grubbs test statistic:
```
> y = c(199.31,199.53,200.19,200.82,201.92,201.95,202.18,245.57)
> 
> grubbs.test(y, type=10)

	Grubbs test for one outlier

data:  y
G = 2.4688000, U = 0.0049308, p-value = 1.501e-07
alternative hypothesis: highest value 245.57 is an outlier
```

However, I'm not sure how to relate the CriticalGrubbs method to other things in R. I see R has the ```qgrubbs``` and ```pgrubbs``` functions. Are those related? Do you know how to use them? I'd like to first better understand how these things work so I can provide more meaningful feedback. I apologize that I have some learning to do here. =P

markrogoyski:
Thinking about what the interface of outlier functions should be, I was thinking, what questions would you want to ask on your data set. I came up with these:

- Does my data contain any outliers?
- How many outliers does it contain?
- What are the outliers?
- Remove outliers from the data.

And specific to Grubb's test, you might want to consider additional questions:

- Is the maximum data point an outlier?
- Is the minimum data point an outlier?
- What is the Grubb's statistic?
- What is the critical value?

Looking at other software to see what they do.
**R - outliers**
- Returns the Grubb's statistic
- Returns the outlier(s) (configurable one/two tails)
- Returns P value

**Python - outlier_utils**
- Can remove outliers from data
- Can return the outlier(s)

Thinking about the kinds of questions you may want to ask, and looking what other libraries provide, what do you think about something like this:

```php
use MathPHP\Statistics\Outlier;

$data = [199.31, 199.53, 200.19, 200.82, 201.92, 201.95, 202.18, 245.57];

// Does data contain outliers?
$boolean = Outlier\Grubbs::hasOutliers($data);
$boolean = Outlier\Grubbs::hasUpperOutlier($data);
$boolean = Outlier\Grubbs::hasLowerOutlier($data);

// What are the outliers?
$arrayOfOutliers = Outlier\Grubbs::getOutliers($data); // Optional second parameter: upper, lower, twoTails

// Remove outliers
$cleanedData = Outlier\Grubbs::removeOutliers($data); // Optional second parameter: upper, lower, twoTails

// Statistic data
$grubbsStatistic = Outlier\Grubbs::statistic($data);
$criticalValue = Outlier\Grubbs::criticalValue($data);
```

Now thinking about these methods, I can immediately think I'd want to use a few of them together. For example, first I ask if there are any outliers, and if so, then I remove them, or get the outliers. So, maybe a Grubbs Test makes sense as an object with the data set as its state in which to call methods on?

Then the code might look like this:
```php
$data = [199.31, 199.53, 200.19, 200.82, 201.92, 201.95, 202.18, 245.57];
$grubbs = new Outlier\Grubbs($test);

// Does data contain outliers?
$boolean = $grubbs->hasOutliers();
$boolean = $grubbs->hasUpperOutlier();
$boolean = $grubbs->hasLowerOutlier();

// What are the outliers?
$arrayOfOutliers = $grubbs->getOutliers(); // Optional parameter: upper, lower, twoTails

// Remove outliers
$cleanedData = $grubbs->removeOutliers(); // Optional parameter: upper, lower, twoTails

// Statistic data
$grubbsStatistic = $grubbs->statistic();
$criticalValue = $grubbs->criticalValue();
```

What do you think are the most common things people would want to do regarding outliers? R and Python seem to have different ideas about this. And neither has the most intuitive interface.

Thoughts?

markrogoyski:
Any update on this? Any thoughts on what an ideal interface would be?

Beakerboy:
Personally, I would just use these functions as-is. However, as the selection of outlier algorithms increases we may want a generalized means of doing some of the things you suggest. For example, a parent outlier class could have some of the methods you suggest, and use the child class to determine what do do for certain common activities. It would be nice to have a "stats person" as a part of the group.

On the topic of groups, have you thought about moving math-php to an organization on Github? If you go to my user page you will see I'm a member of the "flot" organization. It's github's way of adding more complex personnel management to a project. For example, maybe one person can add releases, while another can edit the wiki, or someone can be a member, but have any extra privileges. With the current trend in gameification, there may be a stats person with minor CS experience who would be motivated by being listened to, and having a badge on their profile for being a group member.

markrogoyski:
I'd like to create intention-revealing interfaces that are intuitive to both stats people and programmers that don't necessarily have a math background. I think it is worthwhile to pay attention to how R and other math-centric applications do things, but also consider how those interfaces could be improved using the mechanisms available in an object-oriented language. Math people might not make the best programming design decisions, and programmers might not understand how math people like to work with data. But if they work together, there is a chance to make something great.

For the Grubbs Outlier, the Grubbs object in my second proposal makes sense to me in following the OOP principles of cohesion of data and behavior, and in the sense of having a nice balance of being able to ask questions (hasOutliers) and tell it to do things (removeOutliers). What do you think?

Btw, do you know what R function (or Python function) is similar to what you're calling criticalGrubbs in your pull request? I understand the Grubbs statistic and got that working in R, but I'm not sure how to compute the critical grubbs with R or something else. When I get a pull request, I like to try to understand every line of code. But even if I don't understand the math, at a minimum I'd like to know how to write a unit test using data from an authoritative source, so I at least know how to use it. I'm OK with accepting pull requests with minimal unit tests if I know I can write my own to give confidence that the feature works. So, any help with this would be great. Thanks.

I haven't thought about an organization yet because my main focus is getting to a stable version 1.0. There are a few more features I'd like to add (as well as ask others what features they think absolutely should be in version 1.0), and then I want to spend a lot of time really making sure everything already in the library is rock solid and works. Once that is done, and there is interest to do so, maybe a Github organization is the way to go. I can definitely imagine there will be more maintenance work once version 1.0 is released because I'd like to then move on to 2.0 so we can  move to 7.1 or 7.2 for the main development. But I'd still like to support 1.0 (PHP 7.0) with bug fixes and even still releasing new features in both versions for awhile. Also, as you suggested, if we could get more math people involved, that could generate a lot of good ideas about how to model things and what features are missing, etc. There are a lot of things people could contribute to, so definitely worth thinking about. Thanks for suggesting it.

Beakerboy:
I think the critical value will be returned by `pgrubbs(.05,8,10)`. I made this by following the process laid out by NIST and didn’t look at R, matlab, or numpy.

markrogoyski:
@Beakerboy, 
Are you planning to make some changes and reopen or create a new pull pull request for the Outliers? This is good functionality to have. Let me know if you need any assistance with anything.

Thanks.

Beakerboy:
I think I accidentally clicked close instead of comment. Thanks!

JulienBohy:
Hi.
Any news about This PR ?
I need this for a laboratory project.

markrogoyski:
Hi @JulienBohy,
Thanks for your interest in MathPHP and in this upcoming functionality. No definitive timeline to release Outlier functionality. One option would be to merge Grubs outlier tests soon and then do any sort of larger design work for more Outlier functionality in Version 2.0.

@Beakerboy,
If we put off any larger design and implementation of other Outlier functionality, is the current Grubs functionality ready to go? Also, since MathPHP is still pre-1.0 release, breaking the API is acceptable in semantic versioning if we do want to make any changes, but I like to avoid it if possible. Most people expect composer to update their packages without breaking things in point releases.

Let me know if you want to keep working on this, or if it is good to go as is. If you are happy with it and want to merge it, do you mind creating some more unit tests. Or I can do it. Let me know.

Thanks.

Beakerboy:
The algorithms work fine. Like I said above I’ll leave the larger vision plan to you.

markrogoyski:
OK. Thanks for the input. I'll take a break from the numerical analysis stuff I've been looking at and give this some thought.

@JulienBohy,
We'll try to get it merged and released soon. Thanks for your patience and your feedback.

markrogoyski:
@Beakerboy,
Thanks for working on this. I figured out how to generate test data for all the variations of the Grubbs statistic and critical value to write unit tests against. Everything seems to pass so far. I'll merge this PR as is and add exhaustive unit tests. I'll probably refactor a little bit once I take a look at some other outlier tests and think about it a bit.

Thanks again.

PR #279, opened 2018-11-19
More Efficient Random Number Generation

Beakerboy:


markrogoyski:
Nice. Is this ready to merge, or are there any additional changes coming?

Beakerboy:
There’s really no way to test this so it’s ready to go.If you like this style of random generation there are other algorithms for other distributions that I can add. For example, we can generate random numbers from the T distribution by selecting the appropriate number of random normal numbers and calculating their average z-score.

markrogoyski:
There are tests on the distribution random number functions, but it is mostly just calling them many times and making sure they return a number without crashing. Surprisingly, this caught some bugs in some earlier versions.
If you think the other random number methods are useful, then sure, let's add them.

Thanks!

PR #281, opened 2018-11-30
Mahalanobis Distance

Beakerboy:


Beakerboy:
Yes it looks like that would be better. I’ll change it.  I never noticed the class was there before. Did you ever think that there would be a time where this library had “too much stuff in it” to keep track of?

markrogoyski:
Yeah, the library has grown over the past two years. At the start I thought it would nice to have some descriptive statistics, a normal distribution, and some basic matrix stuff. It just shows that a couple of people taking a few minutes every day, but doing it consistently over the course of multiple years really adds up. Thanks so much for all your contributions!

src/Statistics/Distance.php
```diff
@@ -236,4 +238,41 @@ function ($pᵢ, $qᵢ) {
...
+     *
+     * @param array $x a vector in the vector space. ie [[1],[2],[4]]
+     * @param array $data an array of data. ie [[1,2,3,4],[6,2,8,1],[0,4,8,1]]
+     * @param array $y a vector in the vector space
+     *
+     */
+    public static function Mahalanobis(array $x, array $data, array $y = []): float
+    {
+        $x_matrix = new Matrix($x);
+        $data_matrix  = new Matrix($data);
```
markrogoyski: If the first and only thing that is done with the input parameters is to turn them into Matrixes, would it make sense to just require Matrixes (and a Vector) as the input to the function? Or would this place too much of a burden on the user. Is the expectation that they will more likely have arrays of data rather than a Matrix? Let's do whatever reduces complexity and makes it easiest for the user.

Beakerboy: I am in different about the matrix versus array difference. I chose to go with a raise because if you are pulling data from the database or from some file the first step will be to put it into an array. I figured I would let the function take care of it after that point. However, as I demonstrated in the unit test, dealing with vertical vectors when everything is in an array can be difficult. Having everything in a matrix already would allow the user to transpose data as needed before passing it to the function.

markrogoyski:
Is this pull request ready to be merged? Any other changes coming?

Beakerboy:
I made the submatrix function to simplify the unit testing code. I want to simply extract individual columns from the data matrix to measure their distance. The current selection of function made it difficult. I would have had to pull out a column using `getColumn($x)`, then turn it into a matrix, then transpose it to turn it back into a column matrix. the submatrix would do it in one step.

Beakerboy:
The function is complete. It can calculate the distance between a point and the centroid of a dataset, the distance between two points given the space defined by a dataset, or the distance between the centers of two datasets.

markrogoyski:
Thanks for the updates and additional unit tests.

Just a couple minor things. Add this method to the README.md. Also, add an ```@return``` line to the method to explain what the return value is. Thanks.

src/Statistics/Distance.php
```diff
@@ -236,4 +238,49 @@ function ($pᵢ, $qᵢ) {
...
         return $½D⟮P‖M⟯ + $½D⟮Q‖M⟯;
     }
+
+    /**
+     * Mahalanobis distance
+     *
+     * https://en.wikipedia.org/wiki/Mahalanobis_distance
+     *
+     * The Mahalanobis distance measures the distance between two points in multidimensional
+     * space, scaled by the standard deviation of tje data in each dimension.
```
markrogoyski: Spelling: tje -> the

PR #282, opened 2018-12-04
Submatrix Function

Beakerboy:
Create an arbitrary sized submatrix

### DIFF
--- a/src/LinearAlgebra/Matrix.php
+++ b/src/LinearAlgebra/Matrix.php
@@ -1851,7 +1851,7 @@ class Matrix implements \ArrayAccess, \JsonSerializable
         if ($n₁ >= $this->n || $n₁ < 0 || $n₂ >= $this->n || $n₂ < 0) {
             throw new Exception\MatrixException('Specified Matrix column does not exist');
         }
-        if ($m₂ < $m₁ || $n₂ < $n₁) {
+        if ($m₂ < $m₁) {
             throw new Exception\MatrixException('Ending row must be greater than begining row');
         }
         if ($n₂ < $n₁) {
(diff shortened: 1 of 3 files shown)


src/LinearAlgebra/Matrix.php
```diff
@@ -1830,6 +1831,35 @@ public function adjugate(): Matrix
...
+    {
+        if ($m₁ >= $this->m || $m₁ < 0 || $m₂ >= $this->m || $m₂ < 0) {
+            throw new Exception\MatrixException('Specified Matrix row does not exist');
+        }
+        if ($n₁ >= $this->n || $n₁ < 0 || $n₂ >= $this->n || $n₂ < 0) {
+            throw new Exception\MatrixException('Specified Matrix column does not exist');
+        }
+        $A = [];
+        for ($i = 0; $i <= abs($m₂ - $m₁); $i++) {
+            for ($j = 0; $j <= abs($n₂ - $n₁); $j++) {
```
markrogoyski: Why is the absolute value necessary? Shouldn't the starting row or column always be less than or equal to the ending row or column? Maybe the loop can be simplified by having an extra guard clause on the input parameters prior to entering the loop?

markrogoyski: Add a test case with ending values being smaller than the starting values. What do we expect the result to be?

Beakerboy: I figured that instead of mandating that the "2" indices be larger than the "1" indices, I make it work either way. I'm open to whatever. The way it is currently, these will all return the same thing:
`
$A->submatrix(0,1,2,3);
$A->submatrix(2,1,0,3);
$A->submatrix(2,3,0,1);
$A->submatrix(0,3,2,1);
`

Beakerboy: One of my test cases is already a case with larger indices first.

markrogoyski: It's not like there is some legacy interface requirement that needs to be supported. I think it makes sense to define "normal" ranges as how the function works so it is clear and obvious what it does and what it will return. I'd recommend throwing an exception if the ends are smaller than the starts.

markrogoyski:
Thanks for the pull request to add submatrix!

PR #283, opened 2018-12-13
Fix Divide by zero

Beakerboy:
If two points in the data-set have the same x values, there will be a divide by zero.


markrogoyski:
Thanks for the pull request.

Going forward, when fixing a bug like this, first create a unit test that fails. Then implement the change, and the unit test should pass. This will serve as a regression test going forward to prevent this bug from being reintroduced.

PR #284, opened 2018-12-18
Adding Object Matrix Multiplication

Beakerboy:


src/LinearAlgebra/ObjectSquareMatrix.php
```diff
@@ -106,6 +106,39 @@ public function subtract(Matrix $B): Matrix
...
+        $m = $this->m;
+        $R = [];
+        for ($i = 0; $i < $m; $i++) {
+            for ($j = 0; $j < $n; $j++) {
+                $VA        = $this->getRow($i);
+                $VB        = $B->getColumn($j);
+                $sum = $VA[0]->multiply($VB[0]);
+                foreach ($VA as $key => $value) {
+                    if ($key > 0) {
+                        $sum = $sum->add($VA[$key]->multiply($VB[$key]));
```
markrogoyski: In this case, it seems like the goal is to create a new index in R as the inner product between ```$VA``` and ```$VB```. So rather than mutate ```$sum``` within another nested loop with that awkward conditional after setting it to an initial value, perhaps try to set it in a functional way by doing a map reduce.

```php
$R[$i][$j] = array_reduce(
    array_map(
        function (ObjectArithmetic $a, ObjectArithmetic $b) {
            return $a->multiply($b);
        },
        $VA,
        $VB
    ),
    function ($sum, $item) {
        return $sum
            ? $sum->add($item)
            : $item;
    }
);
```

Take a look at this code now and see what you think about it. As a general rule, we want to reduce complexity as much as possible. Things like conditionals and nested loops, and mutable temporary variables all add to the complexity of trying to reason about code. Functional patterns like map/reduce can usually help in these kinds of cases.

I'll admit that reduce step is not as elegant as you'd expect because we cannot set an initial value, because the ObjectArithmetic interface does not declare a standard constructor.

What do you think?

markrogoyski:
Thanks for the pull request. One code review comment about a possible refactor idea.

Beakerboy:
This is like black magic to me, so I’ll trust you that it is correct. As far as standard constrictors go. I think all number objects can have a “Zero” if that helps. I can make it so that if the constructor is empty, all internal variables are zero.

markrogoyski:
This is one of the great benefits of thorough unit testing! If you have sufficient tests that test outputs based on inputs, then you can freely refactor the internal implementations with confidence with the tests as your guide. I downloaded your branch, made my change and ran the tests and saw green bars.

For the ObjectArithmetic, having all the implementing classes have a _zero_ empty constructor might work here since the reduce step is addition, but if the reduce step was multiplication or something else, the zero value wouldn't work. Reduce allows you to set an initial value for the carry ($sum in our example) that will be used in the first step. When you are reducing a scalar, you can manually set it to 1 or 0 or '' or whatever makes sense for the scalar operation you are working with. In this case, the simplest solution seemed to be just have a simple ternary handle that case within the reducer function.

Anyways, the idea of map reduce is just putting map and reduce together. In math, when you see like  ∑ f(xᵢ), that is basically a kind of map reduce. You map the function f() on to all of the values, and then you reduce all of those function results down to a single value by summing all the values.

markrogoyski:
@Beakerboy 
Is this pull request ready?

Beakerboy:
I’m happy with it.

Issue #286, opened 2019-02-18
Truncate Decimals Frature


specialtactics:
Hi there

I had a requirement in a project recently to be able to truncate decimals from a number without rounding - ie. regardless of whether it would normally be rounded either up or down, the result should be the same - no rounding.

For example, if we wanted to truncate to 2 decimal places, the results for the following 2 numbers would be the same:

1.239 = 1.23
1.231 = 1.23

To my surprise, I couldn't find this sort of functionality in PHP nor any popular maths libraries - because they all insisted on rounding.

So I wrote a function to implement this functionality (along with tests), as it's not very complicated.

I am wondering whether you would accept this sort of thing into this library? I have considered that it can also be said to fit into string manipulation rather than math, but to me and for my purposes, it is strictly a mathematical operation (dealing with financial data).

If you would be happy to have it, I'd be happy to do a PR. Just suggest where I would place such a function, and if you have any name suggestion better than "truncateDecimalNumber", go ahead :)

markrogoyski:
Hi,

Thanks for your interest in MathPHP.

If you think this is a financial-related operation, you could put it in the ```MathPHP\Finance``` class. Otherwise, perhaps the ```MathPHP\Arithmetic``` class would be a good general fit.

What do you think?

specialtactics:
@markrogoyski Sounds good, I will submit a PR for your review

PR #287, opened 2019-02-18
Adds truncateDecimals for arbitrary float

specialtactics:
As described here https://github.com/markrogoyski/math-php/issues/286

### DIFF
--- a/src/Arithmetic.php
+++ b/src/Arithmetic.php
@@ -106,11 +106,11 @@ class Arithmetic
     /**
      * Truncate a given number's decimals, to at most the requested amount of decimal places - without any rounding
      *
-     * @param float $number The number which we are truncating decimals of
+     * @param string $number The number which we are truncating decimals of
      * @param int $places The amount of decimals to keep
      * @return string The result number (Could be a float-like number or not, depending on places > 0)
      */
-    public static function truncateDecimals(float $number, int $places)
+    public static function truncateDecimals(string $number, int $places)
     {
         $positionOfDecimal = strpos($number, '.');
 
@@ -127,7 +127,7 @@ class Arithmetic
         $result = substr($number, 0, $positionOfDecimal) . '.' . $truncatedDecimal;
 
         // If we are truncating to no decimal places, remove the dot at the end
-        if ($places == 0) {
+        if ($places == 0 || strlen($decimal) == 0) {
             $result = substr($result, 0, -1);
         }
 
--- a/src/Arithmetic.php
+++ b/src/Arithmetic.php
@@ -102,4 +102,35 @@ class Arithmetic
             ? abs($magnitude)
             : -abs($magnitude);
     }
+
+    /**
+     * Truncate a given number's decimals, to at most the requested amount of decimal places - without any rounding
+     *
+     * @param float $number The number which we are truncating decimals of
+     * @param int $places The amount of decimals to keep
+     * @return string The result number (Could be a float-like number or not, depending on places > 0)
+     */
+    public static function truncateDecimals(float $number, int $places)
+    {
+        $positionOfDecimal = strpos($number, '.');
+
+        // Get position of decimal - if there aren't any, then result the number
+        if ($positionOfDecimal === false) {
+            return $number;
+        }
+
+        // Get the right number of decimals
+        $decimal = substr($number, $positionOfDecimal + 1);
+        $truncatedDecimal = substr($decimal, 0, $places);
+
+        // Create final result
+        $result = substr($number, 0, $positionOfDecimal) . '.' . $truncatedDecimal;
+
+        // If we are truncating to no decimal places, remove the dot at the end
+        if ($places == 0) {
+            $result = substr($result, 0, -1);
+        }
+
+        return $result;
+    }
 }
--- a/tests/Arithmetic/ArithmeticTest.php
+++ b/tests/Arithmetic/ArithmeticTest.php
@@ -394,13 +394,13 @@ class ArithmeticTest extends \PHPUnit\Framework\TestCase
      * @param        int $places
      * @param        string $expectedResult
      */
-    public function testTruncatesDecimals(float $x, int $places, string $expectedResult)
+    public function testTruncatesDecimals(string $x, int $places, string $expectedResult)
     {
         // When
         $result = Arithmetic::truncateDecimals($x, $places);
 
         // Then
-        $this->assertEquals($expectedResult, $result);
+        $this->assertTrue($expectedResult === $result);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


specialtactics:
@markrogoyski Not sure why it doesn't pick up the test coverage, could you advise ?

markrogoyski:
Hi,
Thanks for the pull request.

I'll take a look at the code coverage and see if I can reproduce it and see what might be the issue.

markrogoyski:
Hi @specialtactics 
I ran the code coverage locally and it showed 100% code coverage.

arithmeticcoverage

I think it must just be an issue with Coveralls. Coveralls also shows some of my existing methods in that class has having no coverage as well. Coveralls has some issue with that class. I'm not sure what it is but I wouldn't worry about it. I'll try some things to see if fixes it.

specialtactics:
Ok, thanks for checking into this @markrogoyski , let me know if there's anything I can do to help.

src/Arithmetic.php
```diff
@@ -102,4 +102,35 @@ public static function copySign(float $magnitude, float $sign): float
...
+    /**
+     * Truncate a given number's decimals, to at most the requested amount of decimal places - without any rounding
+     *
+     * @param float $number The number which we are truncating decimals of
+     * @param int $places The amount of decimals to keep
+     * @return string The result number (Could be a float-like number or not, depending on places > 0)
+     */
+    public static function truncateDecimals(float $number, int $places)
+    {
+        $positionOfDecimal = strpos($number, '.');
```
markrogoyski: I tried this and verified that it works, but I could not find anything in the PHP documentation that states that ```strpos``` has a defined behavior if given a float as the haystack parameter.

PHP documentation: http://php.net/manual/en/function.strpos.php

PHP function declaration:
```php
/**
 * Find the position of the first occurrence of a substring in a string
 * @link https://php.net/manual/en/function.strpos.php
 * @param string $haystack <p>
 * The string to search in
 * </p>
 * @param mixed $needle <p>
 * If <b>needle</b> is not a string, it is converted
 * to an integer and applied as the ordinal value of a character.
 * </p>
 * @param int $offset [optional] <p>
 * If specified, search will start this number of characters counted from
 * the beginning of the string. Unlike {@see strrpos()} and {@see strripos()}, the offset cannot be negative.
 * </p>
 * @return int|boolean <p>
 * Returns the position where the needle exists relative to the beginnning of
 * the <b>haystack</b> string (independent of search direction
 * or offset).
 * Also note that string positions start at 0, and not 1.
 * </p>
 * <p>
 * Returns <b>FALSE</b> if the needle was not found.
 * </p>
 * @since 4.0
 * @since 5.0
 */
function strpos ($haystack, $needle, $offset = 0) {}
```

src/Arithmetic.php
```diff
@@ -102,4 +102,35 @@ public static function copySign(float $magnitude, float $sign): float
             ? abs($magnitude)
             : -abs($magnitude);
     }
+
+    /**
+     * Truncate a given number's decimals, to at most the requested amount of decimal places - without any rounding
+     *
+     * @param float $number The number which we are truncating decimals of
+     * @param int $places The amount of decimals to keep
+     * @return string The result number (Could be a float-like number or not, depending on places > 0)
```
markrogoyski: Should the return value be a string or a float? The reason I ask is because it takes in a float as input. I did some experimenting and it seems like PHP is pretty stable with converting a string to a float and not changing any of the decimal digits. The only exception seemed to be when there were only zeros after the decimal point. In that case, it seemed to truncate everything after the decimal point.

specialtactics: In my local implementation I left it as string, just in case of that edge case as you mentioned, with trailing zeros. Though in theory, that should get removed anyway on it's way in - but just in case, to be correct string-wise.

PHP will happily convert between string and float, even when types are specified.

src/Arithmetic.php
specialtactics: In this case, it will be coerced into string type - documented here;
http://php.net/manual/en/language.types.type-juggling.php

src/Arithmetic.php
markrogoyski: It seems to me it would be more intuitive that if we pass it in a float to truncate, we'd get a float back. So what do you think about changing the return value to float?

specialtactics: @markrogoyski Sounds good, but what about changing param value to a string instead?

markrogoyski: Since we're dealing with Math and Arithmetic here, it seemed (to me at least) the natural inputs and outputs would be numeric values, rather than strings.

Beakerboy:
“Representing numbers as strings in order to guarantee a certain precision or value without flooding point errors”... sounds like a job for [bcmath](
http://php.net/manual/en/intro.bc.php)

src/Arithmetic.php
specialtactics: Hey @markrogoyski I've updated the tests/code to reflect better what it should do (because I think it was mixing string/numeric). Let me know if you think it is still suitable for this lib or not, or if potentially we could make a new namespace for stringified numerical functions.

Otherwise I will close the PR.

markrogoyski: It seems like reasonable functionality that is suitable for MathPHP. I think my preference would be that it work with numbers. It could take a float, do string conversion within its implementation to do the truncation, and then cast back to a float for the return value. I'm probably not understanding the reasons for your preference to keep it a string. Is there a particular scenario where a string would behave differently than a float in terms of the result, given the same input? What is the benefit of string over float here? I'm imagining that most use cases where I have a number with decimals it is a float rather than a string. But maybe that isn't the case?

specialtactics: @markrogoyski the only problem with that approach are trailing zeros. Ie. anytime you have a string with training zeros converted to a float, the trailing zeros will be removed internally in PHP.

So the example is;

123.005 truncated to 2 decimals should be 123.00, however using floats it will return 123.

Why is that important? Maybe you interact with an API which requires a monetary amount figure, and it will expect decimals for those currencies having them, so if you give it 123 it will complain because it expects for the number to have decimals (even if the decimals themselves are 00).

markrogoyski: Sure. That's just how floats work.

I'm just thinking that truncating and representing the float as a string are unrelated problems.

Consider some code like this:

```php
$moneyValueWithUnknownDecimalPrecision = $money1 / $money2;

$moneyValueWithTwoDigitsOfDecimalPrecision = Arithmetic::truncateDecimals($moneyValueWithUnknownDecimalPrecision, 2);

$moneyReport->addMoneyValue(sprintf('%.02f', $moneyValueWithTwoDigitsOfDecimalPrecision));
```

It seems, to me at least, that if keep it general and work with floats, you can easily convert it back to a string with the precision you want. I can understand converting floats to strings for formatting reports in a specific way. My question though, is what is the more general use case? Would the data people are working with usually be numbers or strings? If it is numbers, let's do our best to keep them numbers.

specialtactics: I think it's a good point.

I have actually decided to use a bcmath function (bcsub), because it accomplishes almost exactly what this does, so I will close the PR now.

PR #289, opened 2019-02-27
Whitespace and dev tool upgrades

Beakerboy:
PHPCodeSniffer 3 found this. The version specified in composer.json must not.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
   "require-dev": {
       "phpunit/phpunit": "6.*",
       "php-coveralls/php-coveralls": "^2.0",
-      "squizlabs/php_codesniffer": "2.*",
+      "squizlabs/php_codesniffer": "3.*",
       "phpstan/phpstan": "*",
       "phpmd/phpmd": "^2.6",
       "phploc/phploc": "*"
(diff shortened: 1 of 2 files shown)


markrogoyski:
Thanks for the pull request. Thanks for noticing CodeSniffer has an upgraded version and updating it. Good choice on the PHP versions as well. If a developer has PHP 7.1 or newer than can use PHPUnit 7. But if they are on PHP 7.0, which is the furthest version back MathPHP supports, then they will need PHPUnit 6.

Beakerboy:
There’s a comment buried in the there that math-php fails testing on phpunit 8. It does not like calling a trait from one of our test classes.

markrogoyski:
Yeah. I will take a look. PHPUnit 8 has some other changes and assertion deprecations as well that might need to be addressed. Thanks.

markrogoyski:
@Beakerboy,
I looked into PHPUnit 8. The issue is that ```PHPUnit\Framework\TestCase::setUp()``` is declared with a void return type, and that feature is new with PHP 7.1. Since the current version of MathPHP supports PHP 7.0, the unit tests' setUps cannot be declared void, so they are not compatible with the PHPUnit 8 setUp method. So for now, PHPUnit 8 compatibility is not possible.

MathPHP 1.0 will support PHP 7.0. Version 2.0 will move to either 7.1 or 7.2, and will solve some of these problems.

PR #290, opened 2019-03-18
Qr Decomposition

Beakerboy:
QR Decomposition is the first step towards Eigenvalues for larger matrices, and for more numerically stable Linear Least Squares Regression.

This is an implementation of the Gram-Schmidt process. The Householder process is apparently better, but I do not understand how it works yet.


Beakerboy:
If you approve this and want to get Eigenvalues woking:
````php
public function eigenQR(Matrix $A): array
{
    while(!isTriangular($A)) {
        $QR = $A->GramSchmidtQR();
        $R = $QR['R'];
        $Q = QR['Q'];
        $A = $R->multiply($Q);
    }
    return $A->getDiagonalElements();
}
````

markrogoyski:
Check out the branch qr. I have an implementation of householder reflection. The only issue is that there are multiple QR decompositions and I'm not sure what R and numpy are doing with there signs, so I haven't been able to generate unit tests to validate against yet.

Beakerboy:
Which signs? Is a QR decomposition unique for a given source matrix? If so, you can try the unit tests in this pull request with your QR functions to see if they pass.

markrogoyski:
QR is not unique. There is an arithmetic step where some algorithms say to subtract, some to add, some to do the opposite of the sign of the first element, and then there is whatever R and numpy do.

Beakerboy:
If there’s no consistency, would it make sense to, instead of comparing Q and R to a “known” value, verify that the matrix characteristics are correct, that R is upper triangular, that Q*R=A, and that Q is orthonormal?

markrogoyski:
You can definitely check that QR = A and the upper triangular and orthogonal properties, but I'd do that as an additional unit test. I still want tests that say given some input I expect this output. The question is how do you determine what the output should be. Right now I'm just computing it by hand, but I want to test larger matrices and need an authoritative source to generate my expected values.

If you look online there are a few online calculators and they all get different results. R and NumPy both get the same results. I consider them trusted authoritative sources. And they both most likely are just using Fortran Lapack function calls under the hood.

I haven't yet resorted to trying to understand some old Fortran code to see how they are deciding how to pick the sign of the operation. I was hoping it would be something obvious but I haven't had any luck so far.

Math us usually very precise but QR householder is very inconsistent in how it is described. Even the Wikipedia article on it explains it one way and then the example given doesn't follow their algorithm.

Beakerboy:
I guess we can agree to disagree. If the advantage of Householder is computational, and the the results that the user received are valid, will they care if something went “wrong” and was either done slightly less efficiently? If the code changes, and changes the answer they receive, the unit tests would prove that the new answer is just as good as their old one.

markrogoyski:
It isn't about efficiency. Modifying the matrix in place rather than creating lots of new matrices would be much more efficient, although harder to follow the steps.

The question is does it get a correct answer? Right now I would say "I think so," but I don't have high confidence that it does.

So, I'd like to spend a little more time on it. I welcome any help. If no one can figure it out, then the next best thing will be to check all the properties of Q and R, and things that are built upon them will also test it. It's not my preference, but if we're reasonable confident in it then it's probably OK.

Just an FYI, I code in a TDD method, so I write the tests first. I tend to use R or something to generate my initial test cases. You get immediate feedback as you code. First your test fails. Then you get it to pass. Then you can refactor.

In this case, some of my tests passed, and some of them failed with negative signs in the wrong place. It's been hard to move on to the next step :p

Beakerboy:
I think we differ on the definition of "right answer". For me the "right answer" is any solution that satisfies the properties that one would expect of the decomposition, not "does it agree with vintage fortran code". The algorithm I use is easier to understand, easier to implement, and gives results with satisfy the properties of the decomposition...even if they may not agree with the householder algorithm. I can change my tests to test the properties, and if the algorithm behind the decomposition changes, it will still pass, even if the absolute values in the matrix change.

My coding style is similar..I guess. I find an example, write the function and make sure it "compiles" without tests, then I add one test, fix the errors that come up, then fix the code that causing any failures in the test, then add another test and repeat. I don't use MathPHP to create the tests, it's either from matlab, R, Excel, a textbook, etc.

markrogoyski:
How about we go with your Gram Schmidt implementation for now and build off if that. In the future once I figure out householder better we can swap out the insides and it will all continue to work.

Beakerboy:
So I was just looking at rpubs.com and found the same typo you saw: 

    [6] - sqrt(sum(a))*[1] = 12

Where it should be a plus sign.
I found a typo in another of their pages. The Gram-schmidt lists the three columns of the final Q separately, but changes one element when displaying the final answer. I would call this site untrustworthy.

Beakerboy:
The only thing I can come up with, is for a square matrix, the final H matrix will always be an identity matrix, but with a negative one in the bottom right corner. I'm guessing they skip this last iteration because it will just lead to more math and more floating point errors. The only difference is the signs of the rightmost column of Q will be reversed, as well as the sign on the bottom right element of R.

Beakerboy:
I wrote up my own Householder and got your two unit tests to run. I'm working on non-square matricies.

Beakerboy:
Here is my algorithm. It works on the two unit tests you have in your branch, but fails on a 4x3 matrix.

````php
public function HouseholderQR(): array
    {
        $n = $this->n;  // columns
        $m = $this->m;  // rows
        $HA[-1] = $this;
        $skip_last = $this->isSquare() ? 1 : 0;
        $FullI = MatrixFactory::identity($m);
        for ($i = 0; $i < $n - $skip_last; $i++) {
            $A = $HA[$i - 1]->submatrix($i, $i, $m - 1, $n - 1);
            
            //  The leftmost column of A
            $u = $A->submatrix(0, 0, $A->getM() - 1, 0);
            $uᵀu = $u->transpose()->multiply($u);
            // The square root of the sum of squares
            $α = sqrt($uᵀu[0][0]);
            // We use the sign of the top element of u
            $sgn = Special::sgn($u[0][0]);
            $I = MatrixFactory::identity($A->getM());
            // Get the first column of I
            $e = $I->submatrix(0, 0, $A->getM() - 1, 0);
            $v = $e->scalarMultiply($α * $sgn)->add($u);
            
            // The sum of squares of v
            $vᵀv = $v->transpose()->multiply($v);
            $scalar_vᵀv = $vᵀv[0][0];
            
            $vvᵀ = $v->multiply($v->transpose());
            // We scale $vvᵀ, subtract it from a small Identity, and embed in a large identity
            $H[$i] = $FullI->insert($I->subtract($vvᵀ->scalarMultiply(2 / $scalar_vᵀv)), $i, $i);
            $HA[$i] = $H[$i]->multiply($HA[$i - 1]);
        }
        $R = $HA[$n - 1 - $skip_last];
        $Q = $H[0];
        foreach ($H as $key => $value) {
            if ($key > 0) {
                $Q = $Q->multiply($value);
            }
        }
        return [
            'Q' => $Q,
            'R' => $R,
        ];
    }
````

Beakerboy:
#294 I have not checked it with a source matrix that is wider than it is tall, but it looks good so far.

markrogoyski:
Any update on the QR using Householder reflections? That pull request seems to have been closed.

Beakerboy:
Github says you merged my pull request and that I can delete my branch. (into your qr branch)

PR #292, opened 2019-03-18
Jk EigenValue Algorithm

Beakerboy:
Calculates Eigenvalues of symmetric matricies.


Beakerboy:
Coverage decreased because I have not been able to find a symmetric algorithm that does not converge. We could make "iterations" an optional argument and force it to fail after one iteration.  Alternatively, make the epsilon configurable (via global or function call) and set it excessively low for one unit test.

markrogoyski:
Sure. Go ahead and make the iterations a parameter injectable optional parameter so you can control the behavior and test the scenario where it does not find any eigenvalues.

markrogoyski:
Thanks for continuing to work on this.

I experimented with some test cases and found some issues. Please take a look.

**Test Case - Zero Matrix**

```php
            [
                [
                    [0, 0, 0],
                    [0, 0, 0],
                    [0, 0, 0],
                ],
                [0, 0, 0],
            ],
```

```
There was 1 error:

1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testJKMethod with data set #3 (array(array(0, 0, 0), array(0, 0, 0), array(0, 0, 0)), array(0, 0, 0))
Division by zero

/math-php/src/Functions/Map/Multi.php:126
/math-php/src/LinearAlgebra/Eigenvalue.php:194
/math-php/tests/LinearAlgebra/EigenvalueTest.php:146
```

For reference, here is the result in R:
```
> A = rbind(c(0,0,0), c(0,0,0), c(0,0,0))
> A
     [,1] [,2] [,3]
[1,]    0    0    0
[2,]    0    0    0
[3,]    0    0    0
> eigen(A)
eigen() decomposition
$values
[1] 0 0 0

$vectors
     [,1] [,2] [,3]
[1,]    0    0    1
[2,]    0    1    0
[3,]    1    0    0
```

**Test Case - Ones Matrix**

```php
            [
                [
                    [1, 1, 1],
                    [1, 1, 1],
                    [1, 1, 1],
                ],
                [3, 8.881784e-16, 0],
            ],
```

```
There was 1 error:

1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testJKMethod with data set #3 (array(array(1, 1, 1), array(1, 1, 1), array(1, 1, 1)), array(3, 8.881784E-16, 0))
Division by zero

/math-php/src/Functions/Map/Multi.php:126
/math-php/src/LinearAlgebra/Eigenvalue.php:194
/math-php/tests/LinearAlgebra/EigenvalueTest.php:146
```

Result in R:
```
> A = rbind(c(1,1,1), c(1,1,1), c(1,1,1))
> A
     [,1] [,2] [,3]
[1,]    1    1    1
[2,]    1    1    1
[3,]    1    1    1
> eigen(A)
eigen() decomposition
$values
[1] 3.000000e+00 8.881784e-16 0.000000e+00

$vectors
           [,1]       [,2]       [,3]
[1,] -0.5773503  0.8164966  0.0000000
[2,] -0.5773503 -0.4082483 -0.7071068
[3,] -0.5773503 -0.4082483  0.7071068
```

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -88,4 +93,132 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+        }
+        $m = $A->getM();
+        if ($m < 2) {
+            throw new Exception\BadDataException("Matrix must be 2x2 or larger");
+        }
+
+        $num_zero = 0;
+        $iterationCount = 0;
+        $ε = 1E-16;
+        $Ematrix = [];
```
markrogoyski: I don't see this array variable being used anywhere.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -88,4 +93,132 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+        while ($num_zero < $m * ($m - 1) / 2 && $iterationCount < $iter) {
+            for ($i = 0; $i < $m - 1; $i++) {
+                for ($j = $i + 1; $j < $m; $j++) {
+                    $num = 0;
+                    $den = 0;
```
markrogoyski: $num and $den are redefined a few lines down. These two initialization lines don't seem necessary.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -88,4 +93,132 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+     */
+    public static function JKMethod(Matrix $A, int $iter = 100): array
+    {
+        $originalA = $A;
+        if (!$A->isSymmetric()) {
+            throw new Exception\BadDataException('Matrix must be symmetric');
+        }
+        $m = $A->getM();
+        if ($m < 2) {
+            throw new Exception\BadDataException("Matrix must be 2x2 or larger");
```
markrogoyski: Is there a reason this method can't work on a symmetric 1x1 matrix? I couldn't find a whole lot of information about the JK method online other than some really old PDFs of a Department of Education paper and old Fortran and VB code.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -88,4 +93,132 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+    /**
+     * Replaces a column in a matrix with values from an array
+     *
+     * @param Matrix $matrix
+     * @param array $array of new values
+     * @param int $column
+     *
+     * @returns Matrix with new values in the specified column
+     */
+    private static function replaceColumn(Matrix $matrix, array $array, int $column): Matrix
```
markrogoyski: This function seems like it belongs in the Matrix class.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -88,4 +93,132 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+        for ($i = 0; $i < $m; $i++) {
+            $eigenvalues[] = sqrt(array_sum(Single::square($A->getColumn($i))));
+        }
+        $λ³ = $A
+            ->transpose()
+            ->multiply($originalA)
+            ->multiply($A)
+            ->getDiagonalElements();
+        // The PHP pow() function does not work on negative numbers so
+        // we need to be more complicated.
```
markrogoyski: I had to do something similar for cube root, so I created ```Arithmetic::cubeRoot``` to handle this. Maybe we should have a generalized MathPHP pow function to handle any power. The eigenvalue computation methods should concentrate on what they care about, eigenvalue algorithms, and not quirks of the PHP language.

src/LinearAlgebra/Eigenvalue.php
Beakerboy: Since the algorithm works on pairs of columns, I don't see how it could work on a 1x1. 

The eigenvalue of a 1x1 matrix is just itself, so that's a trivial condition.

src/LinearAlgebra/Eigenvalue.php
Beakerboy: I agree.

src/LinearAlgebra/Eigenvalue.php
Beakerboy: Instead of a `pow()` function, maybe a `root(int $n)` function. `pow()` can accept any real number, and I can understand why it can't work on a negative number, but given an integer, we can easily determine if a root is possible and what the sign will be.

src/LinearAlgebra/Eigenvalue.php
markrogoyski: I think the ```Matrix->eigenvalue``` method should "do the right thing." So, each eigenvalue method should only care about the things it can compute. The Matrix eigenvalue method, however, should know that if it gets a 1x1 or whatever, it does the right thing. And it should probably, by default, always try to do the right thing for the type of Matrix it is, whatever that ends up being.

The optional ```$method``` parameter, then, should probably just be there to override how you want it to behave, if you have reason to do so. Otherwise, it should inspect the Matrix and figure out how to compute the eigenvalues the best way it thinks it can.

src/LinearAlgebra/Eigenvalue.php
Beakerboy: I added an Arthimetic::root() function and a Single::root()

src/LinearAlgebra/Eigenvalue.php
Beakerboy: Function has been moved

src/LinearAlgebra/Eigenvalue.php
Beakerboy: This has been removed

src/LinearAlgebra/Eigenvalue.php
Beakerboy: These have been removed

src/LinearAlgebra/Eigenvalue.php
Beakerboy: I added some decision making code to the Matrix Class. I still think it would be better to recombine Eigenvalue and Eigenvector back into one class, since the two are so interrelated.

Beakerboy:
I believe this most recent version addresses all your concern above. Let me know if there is anything else.

markrogoyski:
Thanks. It's looking good.

I wrote some additional test cases, and ran into a couple of issues. I was able to write one test case that gets incorrect eigenvalues. Some other test cases seem to run out of iterations and not find eigenvalues. I suspect it is an issue with the error tolerance possibly being too small in those cases because it looks like it actually is finding the right values, but cannot meet the conditions to end the iterations in time.

Here is the test case that gets incorrect eigenvalues.
```php
            // Finds incorrect eigenvalues
            [
                [
                    [4, -1, -1, -1],
                    [-1, 4, -1, -1],
                    [-1, -1, 4, -1],
                    [-1, -1, -1, 4],
                ],
                [5, 5, 5, 1],
            ],
```
Test failure output.
```php
There was 1 failure:

1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testJKMethod with data set #12 (array(array(4, -1, -1, -1), array(-1, 4, -1, -1), array(-1, -1, 4, -1), array(-1, -1, -1, 4)), array(5, 5, 5, 1))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 5
-    1 => 5
-    2 => 5
-    3 => 1
+    0 => 4.999999999999999
+    1 => 0.0
+    2 => 0.0
+    3 => 0.0

/math-php/tests/LinearAlgebra/EigenvalueTest.php:160
```
Here is the output in R.
```R
> A = rbind(c(4, -1, -1, -1), c(-1, 4, -1, -1), c(-1, -1, 4, -1), c(-1, -1, -1, 4))
> A
     [,1] [,2] [,3] [,4]
[1,]    4   -1   -1   -1
[2,]   -1    4   -1   -1
[3,]   -1   -1    4   -1
[4,]   -1   -1   -1    4
> eigen(A)
eigen() decomposition
$values
[1] 5 5 5 1

$vectors
           [,1]       [,2]        [,3] [,4]
[1,]  0.8660254  0.0000000  0.00000000 -0.5
[2,] -0.2886751 -0.4345263  0.69126954 -0.5
[3,] -0.2886751 -0.3813938 -0.72194557 -0.5
[4,] -0.2886751  0.8159201  0.03067603 -0.5
```



For your reference, here are all the test cases I added. The four at the end have issues. Everything else passes.

```php
    public function dataProviderForJKEigenvalues(): array
    {
        return [
            [
                [
                    [1, 4],
                    [4, 1],
                ],
                [5.000, -3.000],
            ],
            [
                [
                    [1, 1, 1],
                    [1, 2, 1],
                    [1, 1, 1],
                ],
                [2 + M_SQRT2, 2 - M_SQRT2, 0.00],
            ],
            [
                [
                    [0, 0, 0],
                    [0, 0, 0],
                    [0, 0, 0],
                ],
                [0, 0, 0],
            ],
            [
                [
                    [1, 1, 1],
                    [1, 1, 1],
                    [1, 1, 1],
                ],
                [3, 8.881784e-16, 0],
            ],
            [
                [
                    [4, -30, 60, -35],
                    [-30, 300, -675, 420],
                    [60, -675, 1620, -1050],
                    [-35, 420, -1050, 700],
                ],
                [2585.2538, 37.10149, 1.47805, .166642],
            ],
            [
                [
                    [1, 2],
                    [2, 3],
                ],
                [4.236068, -0.236068],
            ],
            [
                [
                    [1, 2],
                    [2, 1],
                ],
                [3, -1],
            ],
            [
                [
                    [4, 1],
                    [1, -2],
                ],
                [4.162278, -2.162278],
            ],
            [
                [
                    [4, -1],
                    [-1, 9],
                ],
                [9.192582, 3.807418],
            ],
            [
                [
                    [1, 2, 3],
                    [2, 6, 4],
                    [3, 4, 5],
                ],
                [10.784062, 1.825499, -0.609561],
            ],
            [
                [
                    [1, 7, 3],
                    [7, 4, -5],
                    [3, -5, 6],
                ],
                [10.971983, 7.035946, -7.007929],
            ],
            [
                [
                    [5, 6, 7],
                    [6, 3, 2],
                    [7, 2, 1],
                ],
                [13.7082039, -5.0000000, 0.2917961],
            ],
            // Finds incorrect eigenvalues
            [
                [
                    [4, -1, -1, -1],
                    [-1, 4, -1, -1],
                    [-1, -1, 4, -1],
                    [-1, -1, -1, 4],
                ],
                [5, 5, 5, 1],
            ],
            // BadDataException: Eigenvalues not found - iterations run out - maybe the error tolerance is too small?
            [
                [
                    [2, 7, 3],
                    [7, 9, 4],
                    [3, 4, 7],
                ],
                [16.065129, 4.287057, -2.352186],
            ],
            // BadDataException: Eigenvalues not found - iterations run out - maybe the error tolerance is too small?
            [
                [
                    [1, 5, 6, 8],
                    [5, 2, 7, 9],
                    [6, 7, 3, 10],
                    [8, 9, 10, 4],
                ],
                [25.527715, -7.381045, -4.652925, -3.493745],
            ],
            // BadDataException: Eigenvalues not found - iterations run out - maybe the error tolerance is too small?
            [
                [
                    [1, 7, 3, 6],
                    [7, 4, -5, 3],
                    [3, -5, 6, 2],
                    [6, 3, 2, 4],
                ],
                [13.6856756, 9.5813577, -7.2742130, -0.9928203],
            ],
        ];
    }
```

Beakerboy:
I checked your [4,-1,-1,-1] matrix in my VBA implementation of the JK algorithm, and it gives the correct value, so the algorithm works. I must have a mistake somewhere in my PHP implementation.

Beakerboy:
There is something weird going on. I changed a line in Eigenvalue.php from:
`$den = array_sum(Multi::subtract(Single::square($x), Single::square($y)));`
to
`$den = array_sum(Single::square($x)) - array_sum(Single::square($y));`
and it changes to "Eigenvalues not found".

markrogoyski:
I suspect there is some low-level floating point difference and perhaps the error tolerance is too small for every case `$ε = 1E-16;`. Try lowering that error by 2 digits and see if you get a different result.

Maybe the error tolerance needs to be dynamically calculated based on the input values.

Beakerboy:
There is something weird going on. I changed a line in Eigenvalue.php from:
`$den = array_sum(Multi::subtract(Single::square($x), Single::square($y)));`
to
`$den = array_sum(Single::square($x)) - array_sum(Single::square($y));`
and it changes to "Eigenvalues not found".

I think it's floating point issues. When I increased epsilon to 1E-14 it works fine. I'll add your other tests and check it out.

Beakerboy:
I traced the calculations, and PHP is computing a very small negative number for the `$num` variable, while VBA has it as a very small positive number. This is causing the transformation matrix to be the transpose of what it is in VBA, and instead of quickly approaching the solution, it is oscillating and not settling down.

Beakerboy:
I’ve added your additional test cases and increased epsilon. Everything passes.

markrogoyski:
Thanks for continuing to work on this.

I tried some larger matrices and seemed to run into the same problem where epsilon is just slightly too small. Is there some epsilon value that is going to work and get the right results for any generic matrix, or does it need to be dynamically calculated somehow?

Here is the matrix that caused the issue and the result with `$ε = 1E-14`.

PHP unit test data:
```php
            [
                [
                    [-12, -16, -18, 11, 11, 13, -20, 1],
                    [-16, -18, 10, -1, -18, 18, -16, 6],
                    [-18, 10, 4, -17, 2, -14, -11, -16],
                    [11, -1, -17, -1, -19, 5, 8, -20],
                    [11, -18, 2, -19, -13, 8, 5, -4],
                    [13, 18, -14, 5, 8, 10, -19, 19],
                    [-20, -16, -11, 8, 5, -19, 1, -3],
                    [1, 6, -16, -20, -4, 19, -3, 15],
                ],
                [53.85777, -49.65359, -48.21567, 35.73664, -33.18637, 22.86811, 15.47171, -10.87861],
            ],
```

PHPUnit output
```
There was 1 error:

1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testJKMethod with data set #16 (array(array(-12, -16, -18, 11, 11, 13, -20, 1), array(-16, -18, 10, -1, -18, 18, -16, 6), array(-18, 10, 4, -17, 2, -14, -11, -16), array(11, -1, -17, -1, -19, 5, 8, -20), array(11, -18, 2, -19, -13, 8, 5, -4), array(13, 18, -14, 5, 8, 10, -19, 19), array(-20, -16, -11, 8, 5, -19, 1, -3), array(1, 6, -16, -20, -4, 19, -3, 15)), array(53.85777, -49.65359, -48.21567, 35.73664, -33.18637, 22.86811, 15.47171, -10.87861))
MathPHP\Exception\BadDataException: Eigenvalues not found

/math-php/src/LinearAlgebra/Eigenvalue.php:176
/math-php/src/LinearAlgebra/Matrix.php:3698
/math-php/tests/LinearAlgebra/EigenvalueTest.php:146
```

R Reference data
```R
> A
     [,1] [,2] [,3] [,4] [,5] [,6] [,7] [,8]
[1,]  -12  -16  -18   11   11   13  -20    1
[2,]  -16  -18   10   -1  -18   18  -16    6
[3,]  -18   10    4  -17    2  -14  -11  -16
[4,]   11   -1  -17   -1  -19    5    8  -20
[5,]   11  -18    2  -19  -13    8    5   -4
[6,]   13   18  -14    5    8   10  -19   19
[7,]  -20  -16  -11    8    5  -19    1   -3
[8,]    1    6  -16  -20   -4   19   -3   15
> eigen(A)
eigen() decomposition
$values
[1]  53.85777  35.73664  22.86811  15.47171 -10.87861 -33.18637 -48.21567
[8] -49.65359
```

When I changed `$ε = 1E-13` it worked.

Beakerboy:
Google tells me that a 32 bit float has 7 digits of precision, so I guess at every step, epsilon should be something like:
````php
$exp = intval(log10(max(Multi::multiply($x, $y)))) - 7+1;
$ε = 10 ^ $exp;
```
maybe?

If you do a [Error Analysis](https://en.wikipedia.org/wiki/Propagation_of_uncertainty) then 1E-7 * Max(x*y) * sqrt(count(x) * 2) is probably more like it. This works with your new matrix.

markrogoyski:
Interesting. Thanks for investigating this. Would epsilon change for a 64 bit system? Do you need to check `PHP_INT_SIZE` (4 for 32 bit, 8 for 64 bit) before determining what to make epsilon?

Beakerboy:
A 64 bit system would have 1E-15 instead of 1E-7.

Beakerboy:
Need anything else?

markrogoyski:
Looks good. I'll write a few more unit tests just for good measure and then merge presuming they all pass. Sorry for the delay in getting to this.

markrogoyski:
I found a matrix does doesn't seem to get the right eigenvalues.

**PHPUnit test case**
```php
            [
                [
                    [3, 0, 0, 0],
                    [0, 1, 0, 1],
                    [0, 0, 2, 0],
                    [0, 1, 0, 1],
                ],
                [3, 2, 2, 0],
            ],
```

**PHPUnit test run output**
```
There was 1 failure:

1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testJKMethod with data set #22 (array(array(3, 0, 0, 0), array(0, 1, 0, 1), array(0, 0, 2, 0), array(0, 1, 0, 1)), array(3, 2, 2, 0))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 3
-    1 => 2
-    2 => 2
-    3 => 0
+    0 => 3.0
+    1 => 0.0
+    2 => 0.0
+    3 => 0.0

/math-php/tests/LinearAlgebra/EigenvalueTest.php:146
```

**R Output**
```R
> A
     [,1] [,2] [,3] [,4]
[1,]    3    0    0    0
[2,]    0    1    0    1
[3,]    0    0    2    0
[4,]    0    1    0    1
> eigen(A)
eigen() decomposition
$values
[1] 3.000000e+00 2.000000e+00 2.000000e+00 1.110223e-15

$vectors
     [,1] [,2]       [,3]       [,4]
[1,]    1    0  0.0000000  0.0000000
[2,]    0    0 -0.7071068  0.7071068
[3,]    0   -1  0.0000000  0.0000000
[4,]    0    0 -0.7071068 -0.7071068
```

markrogoyski:
What do you think about sorting the return list of eigenvalues. R seems to sort them in descending order.

markrogoyski:
Here is another matrix that does not seem to get the correct eigenvalues.

**PHPUnit test case**
```php
            [
                [
                    [1, 0, 0],
                    [0, 2, 0],
                    [0, 0, 4],
                ],
                [4, 2, 1],
            ],
```

**PHPUnit test run output**
```
There was 1 failure:

1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testJKMethod with data set #25 (array(array(1, 0, 0), array(0, 2, 0), array(0, 0, 4)), array(4, 2, 1))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 4
-    1 => 2
-    2 => 1
+    0 => 0.0
+    1 => 0.0
+    2 => 0.0

/math-php/tests/LinearAlgebra/EigenvalueTest.php:146
```

**R output**
```R
> A
     [,1] [,2] [,3]
[1,]    1    0    0
[2,]    0    2    0
[3,]    0    0    4
> eigen(A)
eigen() decomposition
$values
[1] 4 2 1

$vectors
     [,1] [,2] [,3]
[1,]    0    0    1
[2,]    0    1    0
[3,]    1    0    0
```

Beakerboy:
The eigenvalues of a diagonal matrix are the diagonal elements, so these matrices don’t really need to be solved by the JK method.

markrogoyski:
The first matrix is not diagonal.

Beakerboy:
Yes, sorry, I meant “these” as the diagonal ones.

Beakerboy:
My guess is the fact that columns 2 and 4 are identical is giving some epsilon errors.

Beakerboy:
Closing...jacobi is better

PR #294, opened 2019-03-20
Fixed Householder QR Decomposition

Beakerboy:
The isTriangular() check fails, but the value check passes.


markrogoyski:
IsTriangular compares against 0, and you may have a zero-like value after the arithmetic that isn't quite zero.

Beakerboy:
I commented the check out, re-added your commented out test case, and added one more that I check on r online (https://rdrr.io/snippets/) and it passed.

````

A = matrix(
c(2, 0, 0, 0, -2, -6, 0, 0, -3, -1, 1, 4), # the data elements
nrow=4,              # number of   rows
ncol=3)
qstat = qr(A)
print(A)
print(qr.R(qstat))
````

**Comment by markrogoyski** [comment]
Is this ready to merge? I'll fix/add more unit tests once merged.

**Comment by markrogoyski** [comment]
I tried it with a random 3x4 matrix and it crashed.

FYI here is the PHP unit test data I used:
```php
            [
                [
                    [1, 2, 3, 4],
                    [4, 3, 4, 2],
                    [-3, 6, 7, -3],
                ],
                [
                    'Q' => [
                        [-0.1961161, -0.3096428,  0.9304084],
                        [-0.7844645, -0.5197576, -0.3383303],
                        [0.5883484, -0.7962244, -0.1409710],
                    ],
                    'R' => [
                        [-5.09902,  0.7844645,  0.3922323, -4.1184388],
                        [0.00000, -6.9559051, -8.5815299,  0.1105867],
                        [0.00000,  0.0000000,  0.4511071,  3.4678858],
                    ],
                ],
            ],
```

Crash looks like this:
```
MathPHP\Exception\MatrixException: Specified Matrix row does not exist

math-php/src/LinearAlgebra/Matrix.php:1932
math-php/src/LinearAlgebra/Matrix.php:3402
math-php/tests/LinearAlgebra/MatrixDecompositionsTest.php:2090
```

For reference, here is the same matrix in R computing the QR:
```R
> A = rbind(c(1, 2, 3, 4), c(4, 3, 4, 2), c(-3, 6, 7, -3))
> A
     [,1] [,2] [,3] [,4]
[1,]    1    2    3    4
[2,]    4    3    4    2
[3,]   -3    6    7   -3
> qrA = qr(A)
> qr.Q(qrA)
           [,1]       [,2]       [,3]
[1,] -0.1961161 -0.3096428  0.9304084
[2,] -0.7844645 -0.5197576 -0.3383303
[3,]  0.5883484 -0.7962244 -0.1409710
> qr.R(qrA)
         [,1]       [,2]       [,3]       [,4]
[1,] -5.09902  0.7844645  0.3922323 -4.1184388
[2,]  0.00000 -6.9559051 -8.5815299  0.1105867
[3,]  0.00000  0.0000000  0.4511071  3.4678858
```

Beakerboy:
Fixed. I never tested it on a wide matrix so there were a couple offset errors.

markrogoyski:
Thanks making the changes. After merging, I added more unit tests based on data from R and it's getting some division by zero errors and some failures. I've pushed the latest unit tests to the same **qr** branch. Please run the latest unit tests and take a look. There is probably some edge case in the Householder reflection where uᵀu is zero that needs to be taken care of.

Beakerboy:
I was thinking about this last night and there will be cases where a matrix becomes triangular quite early if there are rows of zeros at the bottom. So the algorithm should be able to cut out at the first moment that it is triangular. This would probably be a better check that the current m-1 adjustment for square and wide matricies. A square Matrix is triangular before it gets to the last reflection so that’s the reason it should stop not because there is anything special about the last reflection.

uᵀu = 0 is certainly a special case. All elements would have to be zero, so I’m guessing the algorithm should just skip that iteration and move on to the next column.

Beakerboy:
I added a check in the householderMatrix function that if alpha == $x[0][0] then just return $I, so that took care of the divide by zero. I also added to the for loop in qrDecompose to kick out if $HA->isUpperTriangular().

However, this means that if you try to qrDecompose an upper triangular matrix, you will get the source matrix and the identity matrix, and this does not seem to be what R does.

Beakerboy:
I've got it down to two failures and no errors. First off, Special::sgn(0) = 0, so that was causing a problem. I had to change it to a ternary `$x[0][0] <0  -1: 1` Then I had it return Identity if uᵀu is zero. I wonder if [[0,1],[0,0]] would be considered a sparse matrix and is using a different algorithm. I don't see how a householder transformation can do anything with this matrix.

markrogoyski:
Cool. Thanks for continuing to investigate and solve these issues. If it turns out that some zero/identity matrices are just edge cases and you can't figure out what R/Fortran is doing, then as long as we know what the edge cases are and how to handle them to get a valid QR result that we are confident in, then I think I'm OK with that, as most of the 'regular' cases are agreeing with R's result. Let's make sure we document the differences where we know they exist either in the code or the unit tests.

When you have the next pull request ready I'll add a bunch more unit tests to that, trying to think of what other edge cases their might be.

Beakerboy:
I'll see if the reddits have any insight, but thanks for catching the divide by zero problems. How was you time away from the computer?

markrogoyski:
With a smartphone you are never really away from a computer. But I tried my best not to look at it too often. I was mostly successful. I haven't yet been tempted to try to write code on my phone. I'm just so used to and prefer PHPStorm or equivalent JetBrains IDE for whatever language I'm using along with my multiple monitors.

Beakerboy:
I probably code as much on my phone as without. Bored waiting for a haircut...github on mobile...

markrogoyski:
The lines at Disneyland were so long, I probably could have gotten a lot done if I had practiced up on coding on my phone =). I need instant feedback from unit tests though. Not sure how I'd pull that off on a phone easily.

markrogoyski:
I've added a Matrix `isOrthogonal` method in the _develop_ branch. You can use this in the QR unit tests. For the identity edge cases where R comes up with different results, maybe it is OK for now to create a separate unit test for those to document them and show that the MathPHP result meets the properties of QR until you have a better idea, if ever, of why R computes a different decomposition.

markrogoyski:
If it helps any, here is the Fortran LINPACK code for QR decomposition using householder transformations.
http://www.netlib.org/linpack/sqrdc.f

I imagine R is probably using this code or something similar. I haven't attempted to try to understand it.  Just thought it might it could be useful if you can't get any answers on the edge cases we've seen.

Beakerboy:
From the comments it sounds like this performs some pivoting steps first. I wonder if this is to optimize the rest of the operations? It might be more efficient to analyze each column and  create a pivot matrix which puts the columns with the most zeros first in order to minimize the number of actual householder transformations which need to be performed

markrogoyski:
I wouldn't worry about optimizations at this point. I like the Kent Beck saying, "Make it work; make it right; make is fast." Let's make it work and make it right. We can optimize later if necessary.

markrogoyski:
If you have it working and it is just the identity edge cases that return a valid QR, just not what R returns, and you are confident that it is correct, go ahead and create a pull request to merge in. I'll write additional unit tests to give confidence that the edge cases work.

PR #310, opened 2019-04-15
MatrixBase Class

Beakerboy:
The first pass at moving universal methods to an abstract base class. Any methods which do not perform arithmetic should be able to move here. The MatrixBase would need to contain or add a abstract function declaration for any `$this->foo()` methods that are used within this class. Adding `implements ObjectArithmetic` to any child class will probably be good too.


markrogoyski:
Thanks for starting to reorganize the Matrix functionality.

I think any type declarations/type hints would be with a MatrixInterface rather than the abstract base class. The abstract base class is more of a hidden implementation detail rather than something to expose.

Therefore, I like the current direction of putting common functionality in the abstract base class. Going forward, think about creating the MatrixInterface so it isn't tempting to put abstract methods with no implementation into the base class. The interface would dictate that a Matrix class must implement those.

Also, is there one MatrixInterface or many interfaces?
* MatrixInterface
* MatrixDataInterface
* MatrixOperationsInterface
* MatrixDecompositionsInterface

I'm not sure the value of more than one interface at the moment. Probably makes sense to to just start with one large one and see how it goes from there.

Beakerboy:
I had not even thought about type hinting the abstract class, so no worries there. I was only planning on having two interfaces, a MatrixInterface and SquareMatrixInterface they would list any methods that are required by the MatrixBase class and SquareMatrixBase trait.

Beakerboy:
Any comments on this work. Is it worth continuing?

markrogoyski:
Looks like there is a return type declaration issue. `Matrix::submatrix` is returning a Matrix instead of a MatrixInterface.

Beakerboy:
I've updated this pull request with the the last 12 months of changes to the Matrix class (thankfully it's pretty stable)

markrogoyski:
I made a couple of new changes to the Matrix class in develop. I'll probably make a couple more as I work out some improvements to address issue #386. Should be minimally impacting what you are doing. But just a heads up. After I release some fix for that I'll take a look at this try to merge it in if you think it is ready to go. Thanks for reorganizing things.

Beakerboy:
Yes, this should be ready to go. I've moved function which operate on the matrix organization to the base class. Any functions that operate on the values within the matrix remain.

src/LinearAlgebra/MatrixBase.php
```diff
@@ -0,0 +1,836 @@
...
+
+        // Same dimensions
+        if ($m != $B->m || $n != $B->n) {
+            return false;
+        }
+
+        // All elements are the same
+        for ($i = 0; $i < $m; $i++) {
+            for ($j = 0; $j < $n; $j++) {
+                if (Support::isNotEqual($this->A[$i][$j], $B[$i][$j], $ε)) {
```
markrogoyski, line 263: I don't think this is going to be safe. `Support::isNotEqual` expects a float and does its own subtraction to compare about the epsilon tolerance. If you did `$objectMatrix->isEqual($anotherObjectMatrix)` this would crash.

src/LinearAlgebra/MatrixBase.php
```diff
@@ -0,0 +1,836 @@
...
+     *         [6 | 3, 4, 5]
+     *
+     * @param  Matrix $B Matrix columns to add to matrix A
+     *
+     * @return Matrix
+     *
+     * @throws Exception\MatrixException if matrices do not have the same number of rows
+     * @throws Exception\IncorrectTypeException
+     */
+    public function augmentLeft(MatrixInterface $B): MatrixInterface
```
markrogoyski, line 389: I'm not sure what's going to happen if you mix two different matrix types here. Suppose you have an ObjectMatrix and you augment it with a NumericMatrix. Probably the objects are going to get cast to floats or something? Otherwise, if somehow a new ObjectMatrix was returned, any arithmetic operation that uses one of the numeric scalars will crash when it does like `$scalarElement->subtract($otherElement)` as the scalar elements are not objects with the `ObjectArithmetic` interface defined on it.

Maybe this needs to have some type checking to make sure only matrices of similar types are augmented.

src/LinearAlgebra/MatrixBase.php
Beakerboy: You're right. The MatrixBase should ensure they are the same type and the same shape. Verifying the equality of each element needs to be left to each child.

Beakerboy: Updated the pull request.

src/LinearAlgebra/MatrixBase.php
Beakerboy: merged in some changes to prevent combining different typed matricies.

Beakerboy:
I've fixed your original concerns, but not I have to figure out why two equal ObjectSquareMatrix objects are not evaluating as equal in the tests,

Beakerboy:
All of the matrix classes affected by this PR are at 100% coverage. The coveralls warning must be dues to fewer overall lines of code.

src/LinearAlgebra/ObjectSquareMatrix.php
```diff
@@ -61,6 +61,38 @@ public function getObjectType(): string
...
+        if (!$this->isEqualSizeAndType($B)) {
+            return false;
+        }
+
+        $m = $this->m;
+        $n = $this->n;
+        // All elements are the same
+        for ($i = 0; $i < $m; $i++) {
+            for ($j = 0; $j < $n; $j++) {
+                if ($this->A[$i][$j] != $B[$i][$j]) {
```
markrogoyski, line 87: The elements of A and B here are of `ObjectArithmetic` right? Maybe this would be safer if that interface defined an `equals` method?

Beakerboy: That’s a good suggestion. I had considered it, but I didn’t want to start messing with that API as a part of this PR. If we follow what Matrix has done, it would be `isEqual()`. It’s not a big change though.

Beakerboy: Hmmm, the classes in the Number namespace have methods called `equals()` while Matrix uses `isEqual()`. If I add a method to the ObjectArithmetic interface, one of these will need to change. Which name do you prefer as a global name?

markrogoyski: `isEquals` usually reads better as a boolean and is consistent with other boolean methods that all start with "is." However, in this case, the ObjectArithmetic interface seems to be just the arithmetic verbs: `add`, `subtract`, `multiply`. So `equals` seems to fit with that pattern. But it is a boolean, and `isEqual` is used everywhere else, so `isEqual` is probably the right choice here.

src/LinearAlgebra/MatrixBase.php
```diff
@@ -0,0 +1,816 @@
...
+     ***************************************************************************/
+
+    /**
+     * Is this matrix the same size and type as some other matrix?
+     *
+     * @param Matrix $B
+     *
+     * @return bool
+     */
+    public function isEqualSizeAndType(MatrixInterface $B): bool
```
markrogoyski, line 249: This could probably be made `protected`. It's an internal implementation detail and probably doesn't have a use to a user of the library or another non-Matrix class.

src/LinearAlgebra/MatrixInterface.php
```diff
@@ -0,0 +1,15 @@
...
+namespace MathPHP\LinearAlgebra;
+
+use MathPHP\Number\ObjectArithmetic;
+
+interface MatrixInterface extends ObjectArithmetic
+{
+    /**
+     * What type of data does the matrix contain
+     *
+     * @return the type of data in the Matrix
```
markrogoyski, line 12: Missing return type: `@return string`.

src/LinearAlgebra/MatrixBase.php
Beakerboy: Done

src/LinearAlgebra/MatrixInterface.php
Beakerboy: Done

src/LinearAlgebra/ObjectSquareMatrix.php
markrogoyski: OK. Is it time to call this PR done and try merging it?

Beakerboy: Yes, good to go if you’re fine with pushing off the `ObjectArithmetic::isEqual()` feature request.

src/LinearAlgebra/Matrix.php
```diff
@@ -1449,7 +1061,7 @@ public function augmentAbove(Matrix $B): Matrix
      * @throws Exception\IncorrectTypeException
      * @throws Exception\MathException
      */
-    public function add(Matrix $B): Matrix
+    public function add($B): Matrix
```
markrogoyski, line 1064: Removing this parameter type declaration will lead to trouble. If you can pass in anything, the next line that starts calling methods on `$B` will fail if it is not a Matrix.

I think the issue is because Matrix no implements ObjectArithmetic that allows mixed parameters. What is the use case for Matrix to implement ObjectArithmetic?

Beakerboy: One future possibility is to enable a matrix of matrices, which is a block matrix. Other than that I don’t think there’s a major advantage other than providing. Unified interface for objects to which algebraic functions can be applied to.

markrogoyski:
Thanks again for continuing to work on this long and complex PR.

I was trying to think about this holistically, and I have some thoughts and questions.

There is the concept of different kinds of matrices, like Numeric, Function, Object, Row and Column ones, etc. And I think you can categorize methods into two general camps:
 - Matrix structure (Operations on rows, columns, number of elements, etc.)
 - Matrix data (Operations on the values of each element)

For structural operations, it doesn't matter what kind of value the matrix holds, as transposing a matrix just means flipping rows and columns, and such. So for these operations, all matrices can share this functionality (Maybe not so for a sparse matrix, but we don't have that yet).

For data operations, adding numeric scalar elements is different than adding Polynomials or Closures because PHP doesn't have a type system where scalars are objects and there is no operator overloading and such. So these types of data operations can't necessarily be shared.

The abstract `MatrixBase` is taking care of the structure operations. This is good and easy to understand.

The trouble I'm running into now, is with the `MatrixInterface`. This interface doesn't say much about the object--just that it can identify its type and it does `ObjectArithmetic`. However, most of the `MatrixBase` operations are now returning a `MatrixInterface`. So if you call `$matrix->augment($someOtherMatrix)`, the object that gets returned isn't a `Matrix`, it is a `MatrixInterface`, but because `MatrixInterface` doesn't define any matrix operations, you don't have any guarantees that you can treat it like a matrix. The same goes for methods that accept a `MatrixInterface` as a parameter rather than a Matrix.

It looks like all the matrix classes still extend `Matrix`, so `Matrix` is still acting as the base class even though it itself extends `MatrixBase`. I don't really see the use of the `MatrixInterface` right now. It creates inconsistencies and allows methods to be called with objects that will not work with the implementation.

There is the other issue, which is not new in this PR, it exists in the current code, where most of the Matrix data methods will fail on some of these non-numeric Matrices. For example: `$polynomialObjectMatrix->negate()` or `$polynomialObjectMatrix->augmentIdentity()` or  `$polynomialObjectMatrix->scalarMultiply`, etc.. Most of these methods will just crash, or produce really strange results, on a non-numeric Matrix.

If my thinking is correct, then any kind of matrix of any data is going to share the structural operations, in which case it could extend `MatrixBase`. But there is more to being a matrix than those operations--there are the data operations as well.

So, let's think about short term and long term plans.

**Short term - this PR in v1.0**
`Matrix` extends `MatrixBase` is a good change to move the common structural code. Maybe there is a better name for this class, but `MatrixBase` works for now as it is abstract and isn't type hinted anywhere.

What do you think about removing `MatrixInterface`, and using `Matrix` as the type parameter and return type for all the methods? This way there is an understanding of what is required and what you get back from the method. `MatrixInterface` doesn't communicate what you actually need or what you have.

Move the `MatrixInterface` `getObjectType` method to `MatrixBase` as an abstract method.

For now `Matrix` would not implement `ObjectArithmetic`

Then merge this PR.

**Medium term - v1.0**
Try to implement the remainder of the Matrix methods for the other kinds of matrices like Object/Function. Maybe some of them just throw exceptions or something. A controlled and documented "not-implemented" or "operation-does-not-apply" exception is better than PHP throwing Notices and Errors and getting strange results.
```php
php > $A = MathPHP\LinearAlgebra\MatrixFactory::create([[new MathPHP\Functions\Polynomial([1, 1])]]);
php > print($A);
[x + 1]
php > $N = $A->negate();
PHP Notice:  Object of class MathPHP\Functions\Polynomial could not be converted to number in /src/LinearAlgebra/Matrix.php on line 1296

php > print($N);
[-1]
```

**Longer term - v2.0**
Time to make backwards-incompatible changes! yay!

Rename `Matrix` to `NumericMatrix`. No matrices other than other numeric matrices (SquareNumericMatrix etc.) would extend it.

Create the interface `Matrix` that all matrix classes must implement. This will include all the matrix operations that are currently defined in the combination of `Matrix` and `MatrixBase`. This is probably a lot of work but at least code doesn't crash in unexpected ways now. All Matrix classes would extend the abstract `MatrixBase` unless there was some reason not to which I'm not sure there would be. `MatrixBase` would not be used as a type hint anywhere--it would just be an implementation detail like it is now.

All parameter types and return types would be the `Matrix` interface for any matrix operation.

There will probably be some object-like matrices that have common methods for implementing many of the matrix functions. Maybe there are common base classes that themselves can extend `MatrixBase`, or maybe there are traits that could work. I'm not so concerned about code reuse if it makes the code more complex. A little code duplication is better than the wrong abstraction.

Maybe figure out how to work in `ObjectArithmetic` to matrix at this time if it makes sense?

Basically implement your design: [Plant UML Matrix Design](http://www.plantuml.com/plantuml/svg/dLHDZXez4Dtx55wmW4_31KQZqV1D56MJ3DAmoiwuAs3Kjeb_511YaMD5sUKeEKdSNNIQIghLAAnClUTNhtxSJp6faBAjHJAf1lbUfUtotLBEcrCm1_bSgnZb6wCsGLaXZ4iGFYaDycMz0vqMmQIjXMJqISZoKrKrdRMdcDSfA9qkVsskazdNn_BttAjqMV3o1XAghevxcCxkPIoWsv0Bhv-p2d2v1eS4heeN5HIfe4lPf0UQhot4ErLJ0Q9wAXVKEhRcfMxBH-1_5K4Qkw_1WajTz8xjHCae9xOxk79G8QYZd2oQMCHyJObhxsjGJhHsNTpxs8PHkqGh4uhuQxOaglAcyK1YfU6sFLz5pq9W1H-oXM0qqpUiwsfwzvHCXicWPBOSEZnLjFOHWdznS9CrGgRM2CfOuz2TDsxW7W2ls0ri6NoLzGZ7Vu70CyQQWk1jFK4ktnB41Xg2eBt55LqZxX7IoL-_4sIFlT4_Kh0FbJ50Kz5Rfzxuisz81w96kupEj7Uxl549Llrh56SgmbPF4mnKByXqMHs4nfhO5_OsEvsCTuEh8ickK3hT5Wfp2qFXaPMX_fXYU_1VTmPp47APcqaF3tAwklD1Rqqrau-Fg40iJ1O_ltx_yQq5-E-nbELGbLgMmJXjzcMhZFEX8I7TNZMfGL2vntz_Wxcb_My-_zktOBl5KzbL6kd0K1y7HFZC3Bib3eW8-ch6-K_WgcphNm00)

But the `Matrix` interface would have all the matrix interface methods in it and try to only use that in parameter and return type hints.

---

Thoughts?

By the way, thanks again for continuing to work on this and improve what has become a large matrix codebase. I think we are making progression on making sure we have the right model and that things work, regardless of what matrix you have, and making the codebase as clean as possible.

Beakerboy:
We've had some of these conversations in the Issues queue. Some of the changes that have been implemented here are in preparation for the longer term goals. We both agree that MatrixBase is not the greatest. Long term, this could be Matrix, and Matrix will become NumericMatrix. Since the MatrixBase is abstract, users will likely never encounter it, so the naming really is not too important.

ObjectSquareMatrix will not extend from NumericMatrix in the long term, so this could be decoupled. Only functions that return something meaningful could be implemented instead of having to test for those conditions.

markrogoyski:
I merged it! This might take the record of both largest pull request and longest open that was merged (a year and a half!).

Thanks for working on this.

PR #314, opened 2019-05-03
just a ; on the end

tikky:
small typing mistake. It was ":" instead of ";"


markrogoyski:
Thank you for your interest in MathPHP and wanting to improve it!

PR #317, opened 2019-05-24
IsNormal matrix

Beakerboy:
If a matrix is normal, it is diagonalizable, and has eigenvalues and eigenvectors

### DIFF
--- a/src/LinearAlgebra/Matrix.php
+++ b/src/LinearAlgebra/Matrix.php
@@ -1052,7 +1052,7 @@ class Matrix implements \ArrayAccess, \JsonSerializable
             return false;
         }
 
-        // AAᵀ = I
+        // AAᵀ = AᵀA
         $I   = MatrixFactory::identity($this->m);
         $Aᵀ  = $this->transpose();
         $AAᵀ = $this->multiply($Aᵀ);
--- a/src/LinearAlgebra/Matrix.php
+++ b/src/LinearAlgebra/Matrix.php
@@ -1053,7 +1053,6 @@ class Matrix implements \ArrayAccess, \JsonSerializable
         }
 
         // AAᵀ = AᵀA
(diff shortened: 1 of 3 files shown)


markrogoyski:
Thanks for the pull request. Nice addition.

Just curious. Is it expected that this will just work with an ObjectMatrix of Complex numbers? I think it would need to check the conjugate transpose of that. I didn't see any complex number unit tests, and I presume complex matrices are just not a public-facing feature yet, which is fine. I want to make sure I have the same understanding you do so I don't write some unit tests for something that isn't expected to work yet.

Thanks!

Beakerboy:
Eventually, sure. Since we don’t have a ComplexNumber Class yet I didn’t write any tests for that. I’ve been treating the existing Matrix class as a NumericalMatrix. When the Matrix refactor goes through, i’ll spend some time either abstracting functions to work in different contexts, or writing different versions for each class.

markrogoyski:
Cool. I just wanted to make sure we had the same understanding. I like that thought, that the Matrix is the NumericMatrix. Makes it easy. Thanks.

src/LinearAlgebra/Matrix.php
```diff
@@ -1035,6 +1036,31 @@ public function isOrthogonal(): bool
...
+     * @throws Exception\MathException
+     */
+    public function isNormal(): bool
+    {
+        if (!$this->isSquare()) {
+            return false;
+        }
+
+        // AAᵀ = AᵀA
+        $I   = MatrixFactory::identity($this->m);
```
markrogoyski: Why do you need the identity matrix here? Doesn't look like it is used.

Beakerboy: Sorry, I just copied the function above and (poorly) modified it. An orthogonal Matrix is a special type of normal matrix.

PR #325, opened 2019-08-06
Power iteration Eigenvalue calculation

Beakerboy:
This is another algorithm that easily calculates both eigenvalues and eigenvectors, so It might be worthwhile considering how to combine the two static classes into one.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
       "phpstan/phpstan": "*",
       "phpmd/phpmd": "^2.6",
       "phploc/phploc": "*",
-      "php-mock/php-mock-phpunit": "*"
+      "php-mock/php-mock-phpunit": "2.*"
   },
   "autoload-dev": {
     "psr-4": { "MathPHP\\Tests\\": "tests/" }
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
       "squizlabs/php_codesniffer": "3.*",
       "phpstan/phpstan": "*",
       "phpmd/phpmd": "^2.6",
-      "phploc/phploc": "*"
+      "phploc/phploc": "*",
+      "php-mock/php-mock-phpunit": "*"
   },
   "autoload-dev": {
     "psr-4": { "MathPHP\\Tests\\": "tests/" }
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -120,8 +120,8 @@ class Eigenvalue
         for ($i = 0; $i <$m; $i++) {
             $random_array[] = [rand()];
         }
-        $b = MatrixFactory::create($random_array);
-        $new_b = MatrixFactory::zero($m, 1);
+        $new_b = MatrixFactory::create($random_array);
+        $b = MatrixFactory::zero($m, 1);
         $abs = -1;
         $new_abs = -2;
         while (!$b->isEqual($new_b)) {
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -33,7 +33,7 @@ class Eigenvalue
      *
      * @throws Exception\BadDataException if the matrix is not square
      */
-    private static checkMatrix(Matrix $A)
+    private static function checkMatrix(Matrix $A)
     {
         if (!$A->isSquare()) {
             throw new Exception\BadDataException('Matrix must be square');
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -132,15 +132,9 @@ class Eigenvalue
     {
         self::checkMatrix($A);
         
-        $m = $A->getM();
-        $random_array = [];
-        for ($i = 0; $i <$m; $i++) {
-            $random_array[] = [rand()];
-        }
-        $b = MatrixFactory::create($random_array);
+        $b = MatrixFactory::random($A->getM(), 1);
         $newμ = 0;
         $μ = -1;
-        $new_abs = -2;
         while (!Support::isEqual($μ, $newμ)) {
             if ($iterations <= 0) {
                 throw new Exception\FunctionFailedToConvergeException("Maximum number of iterations excecuted.");
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -3,6 +3,7 @@ namespace MathPHP\LinearAlgebra;
 
 use MathPHP\Exception;
 use MathPHP\Functions\Polynomial;
+use MathPHP\Functions\Support;
 
 class Eigenvalue
 {
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -123,10 +123,10 @@ class Eigenvalue
      * https://en.wikipedia.org/wiki/Power_iteration
      * @param Matrix $A
      *
-     * @return float most extreme eigenvalue
+     * @return float[] most extreme eigenvalue
      * @throws Exception\BadDataException if the matrix is not square
      */
-    public static function powerIteration(Matrix $A): float
+    public static function powerIteration(Matrix $A): array
     {
         self::checkMatrix($A);
         
@@ -148,6 +148,6 @@ class Eigenvalue
             $newμ = $new_b->transpose()->multiply($A)->multiply($new_b)->get(0, 0) / $new_b->transpose()->multiply($new_b)->get(0, 0);
         }
         
-        return $newμ;
+        return [$newμ];
     }
 }
--- a/src/LinearAlgebra/Eigenvalue.php
+++ b/src/LinearAlgebra/Eigenvalue.php
@@ -145,7 +145,7 @@ class Eigenvalue
             $μ = $newμ;
             $Ab = $A->multiply($b);
             $new_b = $Ab->scalarDivide($Ab->frobeniusNorm());
-            $newμ = $new_b->transpose()->multiply($A)->multiply($new_b) / $new_b->transpose()->multiply($new_b);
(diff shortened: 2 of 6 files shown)


markrogoyski:
Thanks for the pull request that provides another good eigenvalue method.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -89,4 +104,49 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+        $m = $A->getM();
+        $random_array = [];
+        for ($i = 0; $i <$m; $i++) {
+            $random_array[] = [rand()];
+        }
+        $b = MatrixFactory::create($random_array);
+        $newμ = 0;
+        $μ = -1;
+        $new_abs = -2;
+        while (!Support::isEqual($μ, $newμ)) {
```
markrogoyski: Does there need to be some alternate max iterations condition or something just as a check to make sure we don't go into an infinite loop on some edge case?

tests/LinearAlgebra/EigenvalueTest.php
```diff
@@ -20,6 +20,21 @@ public function testClosedFormPolynomialRootMethod(array $A, array $S)
...
+     * @testCase     powerIterationMethod returns the expected eigenvalues
+     * @dataProvider dataProviderForEigenvalues
+     * @param        array $A
+     * @param        array $S
+     */
+    public function testPowerIteration(array $A, array $S)
+    {
+        $A = MatrixFactory::create($A);
+        // Find the most extreme value
+        $eigenvalue = max($S) ** 2 > min($S) ** 2 ? max($S) : min($S);
```
markrogoyski: We write unit tests to validate our code that is doing complex computations. Generally you don't want the test code itself doing complicated computations which could introduce mistakes into the test. For this test, it is probably best to just add a third parameter to the data provider that represents the greatest absolute value eigenvalue.

I was thinking, maybe this would be a useful function to write, `absoluteMax`. Give it an array of numbers and it finds the value with the greatest absolute value. Would something like that be generally useful? And if so, what would it return for a list like [-4, 1, 3, 4]? Would you get 4 or -4?

Beakerboy: I was thinking that, but I feel that for an arbitrarily simple expression like this, it makes more sense to have data providers that can be used in multiple tests, then to add in extra fluff that is needed for this one test. As other eigenvalue methods come into the class, any specific test data made for them can all be passed through the powerIteration function.

markrogoyski: Yeah, I actually started writing that method, and then when I wrote my first couple test cases, I kind of got stuck.

For the data provider, just FYI, your data provider could have 10 parameters, but if you test only uses the first two, your test method declaration only needs to declare the first two. This is fine.

Your data provider could look like this:
```php
    public function dataProviderForEigenvalues(): array
    {
        return [
            [
                [
                    [0, 1],
                    [-2, -3],
                ],
                [-2, -1],
                -2
            ],
        ];
    }
```

And you can have two tests using different elements:
```php
    public function testClosedFormPolynomialRootMethod(array $A, array $S)
    {
        // ...
    }

    public function testPowerIteration(array $A, array $S, int $expectedGreatestAbsoluteEigenvalue)
    {
        // ...
    {
```

Consider if this kind of generic, yet multi-purpose kind of data provider would be useful here.

Beakerboy: Ahh, that's better than I was assuming, but having to include `array $S` as a pointless function parameter still bugs me. I'll do it.

src/LinearAlgebra/Eigenvalue.php
Beakerboy: I added a new exception and updated this.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -89,4 +104,54 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+     *
+     * will converge to the dominant eigenvector,
+     *
+     * The corresponding eigenvalue is calculated as:
+     *
+     *      bₖᐪAbₖ
+     * μₖ = -------
+     *       bₖᐪbₖ
+     *
+     * https://en.wikipedia.org/wiki/Power_itera
```
markrogoyski: Looks like a copy and paste mistake or something on the end of this URL.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -89,4 +104,54 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+     * @param Matrix $Ation
+     * @param int $iterations max number of iterations to perform
+     *
+     * @return float[] most extreme eigenvalue
```
markrogoyski: Since this only returns a single eigenvalue, would it make more sense to just return the value and not wrap it in an array? The `Matrix::eigenvalues` has to return an array, but this method doesn't seem to need to return an array. However, if this returns just the value, then the `Matrix::eigenvalues` needs to know what to do with each method and how to handle the response to provide an array of eigenvalues per its contract. To simplify this interface, `Matrix::eigenvalues` needs to add some complexity to handle this simplification. What do you think? I tend to lean towards simple public interfaces even at the expense of complexity in the implementations.

src/LinearAlgebra/Eigenvalue.php
Beakerboy: Fixed

src/LinearAlgebra/Eigenvalue.php
Beakerboy: My thought was to keep the return type the same as other implementations. Just like how the result of a 1x1 Matrix will be a single value from any other algorithm, but we would still return it as an array.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -89,4 +104,54 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+    public static function powerIteration(Matrix $A, int $iterations = 1000): array
+    {
+        self::checkMatrix($A);
+        
+        $m = $A->getM();
+        $random_array = [];
+        for ($i = 0; $i <$m; $i++) {
+            $random_array[] = [rand()];
+        }
+        $b = MatrixFactory::create($random_array);
```
markrogoyski: How about replacing all of this code with a new factory method that describes what we want?
```php
$b = MatrixFactory::random($A->getM(), 1);
```

We want a random m x 1 matrix to initialize the power iteration. The work to create this matrix is not relevant to this algorithm and distracts from what is the focus here. So let's move this responsibility somewhere else that cares about creating different kinds of Matrices.

Beakerboy: done

tests/LinearAlgebra/MatrixFactoryTest.php
```diff
@@ -837,4 +838,17 @@ public function testHilbertExceptionNLessThanZero()
...
+
+    /**
+     * Test that the appropriate matrix is constructed
+     *
+     * @runInSeparateProcess
+     */
+    public function testRandomMatrix()
+    {
+        $expected = [[.31415926535]];
+        $rand = $this->getFunctionMock('MathPHP\LinearAlgebra', 'rand');
```
markrogoyski: Neat!

tests/LinearAlgebra/MatrixFactoryTest.php
```diff
@@ -837,4 +838,17 @@ public function testHilbertExceptionNLessThanZero()
         $this->expectException(Exception\OutOfBoundsException::class);
         MatrixFactory::hilbert(-1);
     }
+
+    /**
+     * Test that the appropriate matrix is constructed
+     *
+     * @runInSeparateProcess
```
markrogoyski: Is this necessary? I looked through PHPMock and it looks like when you call `getFunctionMock` it registers the FunctionMock as a listener on the test result. At the end of the test, it will call each listeners' `endTest` method, of which the PHPMock's MockDisabler will deactivate the mock. Try it without this annotation and see if it works.

Beakerboy: I had issues in another project without this. If rand() is called before the FunctionMock is registered, calls to rand() will go to \rand() despite the existence of a namspaced function. So if any test in the MathPHP\LinearAlgebra namespace calls rand(), this will be needed. I'll check...

composer.json
```diff
@@ -16,7 +16,8 @@
       "squizlabs/php_codesniffer": "3.*",
       "phpstan/phpstan": "*",
       "phpmd/phpmd": "^2.6",
-      "phploc/phploc": "*"
+      "phploc/phploc": "*",
+      "php-mock/php-mock-phpunit": "*"
```
markrogoyski: Probably best to be specific with the release. Use 2.4 or 2.*.

Beakerboy: Got it. I was thinking it defaulted to most recent acceptable version.

Beakerboy:
the @runInSeparateProcess annotation IS needed. Also I've been meaning to ask, the @testCase annotaion is not an official PHPUnit annotation. is there some other purpose for it?

markrogoyski:
> the @runInSeparateProcess annotation IS needed. Also I've been meaning to ask, the @testcase annotaion is not an official PHPUnit annotation. is there some other purpose for it?

It was a mistake. A while ago I updated the coding standards: https://github.com/markrogoyski/math-php/wiki/Coding-Standards#unit-tests

It should have been `@test` and not `testcase`. The reason to add the `@test` annotations is like a safety net. PHPUnit will run any method that starts with the word "test" or has the @test annotation. In the past, multiple times (on this project and others) there were unit tests that were named improperly and never run. So, adding the annotation is like a second chance to get it right, to make sure PHPUnit runs the test.

tests/LinearAlgebra/MatrixFactoryTest.php
markrogoyski: OK. Thanks for investigating.

markrogoyski:
Thanks for working on the power iteration eigenvalue method.

PR #326, opened 2019-08-08
Jacobi

Beakerboy:
Here is another eigenvalue algorithm. This can replace the JK method that's been in the works for a while. I finally figured out how this works and implemented it. I also added in all the test cases that you had made for JK and they all pass.


src/LinearAlgebra/Eigenvalue.php
```diff
@@ -89,4 +92,108 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+    * https://en.wikipedia.org/wiki/Givens_rotation
+    *
+    * @param int $i The row in G in which the top of the roatation lies
+    * @param int $j The column in G in which the left of the roatation lies
+    * @param float $angle The angle to use in the trigonometric functions
+    * @param int $m The total number of rows in G
+    *
+    * @return Matrix
+    */
+    private static function givensMatrix(int $i, int $j, float $angle, int $m) : Matrix
```
markrogoyski: Would this make more sense to be in the `MatrixFactory` class?

Beakerboy: I don’t know how much use it would have outside this class. I guess it couldn’t hurt.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -89,4 +93,78 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+    }
+
+    /**
+     * True if all off-diagonal elements are very close to zero
+     *
+     * @param Matrix $A
+     *
+     * @return bool
+     */
+    private static function isNearlyDiagonal(Matrix $A): bool
```
markrogoyski: Would it be better to update the Matrix class's `isDiagonal` and `isTriangular` methods to use the approximately zero Support methods instead of defining this helper class which is doing that?

Beakerboy: If that's the default behavior that you prefer, then I'm good with it.

markrogoyski:
Looks like some merge conflicts arose after the last PR merge. Should be fairly straightforward to resolve.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -106,6 +109,60 @@ public static function closedFormPolynomialRootMethod(Matrix $A): array
...
+                        $pivot['value'] = $D[$i][$j];
+                        $pivot['i'] = $i;
+                        $pivot['j'] = $j;
+                    }
+                }
+            }
+            $i = $pivot['i'];
+            $j = $pivot['j'];
+            if ($D[$i][$i] == $D[$j][$j]) {
+                $angle = ($D[$i][$i] <=> 0) * \M_PI / 4;
```
markrogoyski: I just want to verify the intent here.
- If the element is 0, you want and angle of 0
- If the element is a negative number, you want an angle of -pi/4
- If the element is a positive number, you want an angle pi/4

Correct?

Beakerboy: It looks like the zero case is not correct.

Beakerboy:
When this is merged, I can very quickly submit a PCA class.

PR #328, opened 2019-08-09
unnecessary Condition

Beakerboy:
The householder function has an unnecessary condition that is not tested. I had this in there when trying to figure out the best way to remove the divide by zero error.


Beakerboy:
This is causing the two other pull request to give "decreased coverage" warnings.

PR #331, opened 2019-08-13
Update Eigenvector.php

Beakerboy:
slight refactor


markrogoyski:
Thanks for the pull request and the improvements. Sorry it took a bit to get to this. I was camping in Yosemite without cell signal for a few days.

Beakerboy:
It’s never a problem. Sounds fun..I should get to bed.

PR #332, opened 2019-08-13
PCA

Beakerboy:
* Created PCA Class
* Added Sample Data
* Created Unit Tests


Beakerboy:
Let me know if you think this belongs in a different namespace

markrogoyski:
Thanks for the pull request implementing some of your must-have V1 features. Give me a few days to become familiar with PCA so I can do a meaningful code review. Thanks.

src/LinearAlgebra/PCA.php
```diff
@@ -0,0 +1,323 @@
+<?php
+
+namespace MathPHP\LinearAlgebra;
```
markrogoyski: What do you think about moving this to a new namespace: `MathPHP\Statistics\Multivariate`?

Beakerboy: Sure. I need to add a few other test to cover code sections where the PCA model is applied to new data. I’ll work on it this weekend while my family is on vacation without me 😥

markrogoyski:
Is this naming convention standard?
https://en.wikipedia.org/wiki/Principal_component_analysis#Table_of_symbols_and_abbreviations

Should we use the same names for the variables? I like mathematical variable names, but I also like code to be understandable. I wouldn't know what these mean, but would someone reasonably familiar with PCA know them?

tests/Data/SampleData.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace MathPHP\Tests\Data;
+
+class SampleData
+{
+    public static function mtcars()
```
markrogoyski: This is great that you've used a standard dataset and have all the cross checking with R for the unit tests!

src/Statistics/Multivariate/PCA.php
```diff
@@ -0,0 +1,324 @@
...
+        return $result_matrix;
+    }
+
+    /**
+     * Calculate the critical limits of T²
+     *
+     * @param float $alpha the probability limit of the critical value
+     * @return float[] Critical values for each model complexity
+     */
+    public function getCriticalT²(float $alpha = .05): array
```
markrogoyski: For internal variable names and methods I like using fancy characters, but for a public interface method, my preference is to stick to alphanumeric characters. In my PHPStorm IDE I don't have to do much to get it to auto-complete `getCriticalT²` for me, but I'm not sure what someone using vim or something would do.

Another option, is to make a wrapper method, so you'd have both `getCriticalT²` and `getCriticalT2` so you can use one internally and one for the public interface if you still want to keep the fancy version.

tests/Statistics/Multivariate/PCATest.php
```diff
@@ -0,0 +1,381 @@
...
+        $loadings = $this->pca->getLoadings();
+        $load_array = $loadings->getMatrix();
+        
+        //Get an array that's roughly ones and negative ones.
+        $quotiant = Multi::divide($expected[1], $load_array[1]);
+        
+        // Convert to exactly one or negative one. Cannot be zero.
+        $signum = array_map(
+            function ($x) {
+                return $x <=> 0;
```
markrogoyski: This line doesn't match the comment above. If `$x` is 0 then this will return 0.

Beakerboy:
> Is this naming convention standard?
> https://en.wikipedia.org/wiki/Principal_component_analysis#Table_of_symbols_and_abbreviations
> 
> Should we use the same names for the variables? I like mathematical variable names, but I also like code to be understandable. I wouldn't know what these mean, but would someone reasonably familiar with PCA know them?

It looks understandable. If you wanted me to, I could change names and add a comment to the initial assignment verifying what the variable is.

src/Statistics/Multivariate/PCA.php
Beakerboy: Less fancy is fine.

tests/Statistics/Multivariate/PCATest.php
Beakerboy: $x cannot be zero, because it is the result of a division, and the denominator would have to be infinitely larger than the numerator.

markrogoyski:
This looks great. Are you planning to add anything else or is this ready to go? Your call if you want to rename the variables or not. I don't have a strong opinion on that. Readability and understandability is what is important.

markrogoyski:
Anything left to do on this? Is this ready to merge?

Beakerboy:
I think it's ready.

markrogoyski:
I realized that there are not unit tests for the three other scenarios of center and scale not both being true. I started writing unit tests for these. I started with center=FALSE and scale=TRUE. For the most part I was able to figure out how to do this in R and create unit test data. However, there are two tests that I can't seem to get to pass. Do you mind taking a look.

In branch https://github.com/markrogoyski/math-php/tree/pca-unit-tests
I have a new test file https://github.com/markrogoyski/math-php/blob/pca-unit-tests/tests/Statistics/Multivariate/PCACenterFalseScaleTrueTest.php

I'm having difficulty with the ```testEigenvalues``` and ```testCriticalQ``` tests. I'm not sure if I'm doing something wrong in R or if there is a bug.

Also, there were two places in the PCA class where the PHP function ```array_fill``` was used without a value. Those code paths weren't being executed with center and scale both true, but now they are. I put in 0. Was that your intention?

Other than the two tests I mentioned, everything else seems to work for center=FALSE and scale=TRUE.

Any help would be appreciated. Thanks.

src/Statistics/Multivariate/PCA.php
```diff
@@ -0,0 +1,317 @@
...
+    public function __construct(Matrix $M, bool $center = true, bool $scale = true)
+    {
+        // Check that there is enough data: at least two columns and rows
+        if (!($M->getM() > 1) || !($M->getN() > 1)) {
+            throw new Exception\BadDataException('Data matrix must be at least 2x2.');
+        }
+        if ($center === true) {
+            $this->center = $M->columnMeans();
+        } else {
+            $this->center = new Vector(array_fill(0, $this->data->getN()));
```
markrogoyski: This and also line 73 are missing the value.

src/Statistics/Multivariate/PCA.php
```diff
@@ -0,0 +1,317 @@
...
+        $vars = $this->data->getN();
+        $Qcrit = [];
+        for ($i = 0; $i < $vars - 1; $i++) {
+            $evals = array_slice($this->EVal->getVector(), $i + 1);
+            $t1 = array_sum($evals);
+            $t2 = array_sum(Single::square($evals));
+            $t3 = array_sum(Single::pow($evals, 3));
+            $h0 = 1 - 2 * $t1 * $t3 / 3 / $t2 ** 2;
+            if ($h0 < .001) {
+                $h0 = .001;
```
markrogoyski: Can you document the purpose of this conditional and give some guidance on what kind of data would trigger it so I can write a unit test for it?

Beakerboy: The code was pretty much directly lifted from R. I imagine the point is it’s the best fit for this empirical formula, to keep it from blowing up at low h0 when 1 / $h0 is used as an exponent. I think it should be able to be triggered when scaling is not used to standardize the data, and all the data is very small.

PR #335, opened 2019-08-14
Matrix isIdempotent property

Beakerboy:
Test if A*A=A

### DIFF
--- a/src/LinearAlgebra/Matrix.php
+++ b/src/LinearAlgebra/Matrix.php
@@ -819,7 +819,7 @@ class Matrix implements \ArrayAccess, \JsonSerializable
     public function isIdempotent(): bool
     {
         $A² = $this->multiply($this);
-        return $A²->getMatrix() == $this->A;
+        return $this->isEqual($A²);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/LinearAlgebra/Matrix.php
```diff
@@ -808,6 +809,19 @@ public function isRref(): bool
...
+     * Is the matrix idempotent?
+     * A matrix that equals itself when squared.
+     * https://en.wikipedia.org/wiki/Idempotent_matrix
+     *
+     * @return boolean true if matrix is idempotent; false otherwise
+     */
+    public function isIdempotent(): bool
+    {
+        $A² = $this->multiply($this);
+        return $A²->getMatrix() == $this->A;
```
markrogoyski: This is not a safe way to compare two matrices, because one could some property like `$ref` set and the new one would not, and therefore the objects would not be equal. I recommend using the `isEqual` comparison method.

Beakerboy: I just copied the method below it `isInvolutory()`, so it has the same problem. I don’t know who coded that one up though.

markrogoyski: Probably me! I definitely need to update some things based on some new ways of doing things. Either way, we don't want to spread the problem into new code. =)

Beakerboy: It turns out it isn’t actually a problem because the comparison is between the arrays, not the entire objects.

markrogoyski:
Nice addition. Thanks for the pull request.

PR #337, opened 2019-08-16
Moved LU Decomposition to separate class and new namespace

Beakerboy:
This change is internal-only; it does not affect the user.

### DIFF
--- a/src/LinearAlgebra/Decomposition/LU.php
+++ b/src/LinearAlgebra/Decomposition/LU.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace MathPHP\LinearAlgebra\Decompositions;
+namespace MathPHP\LinearAlgebra\Decomposition;
 
 use MathPHP\Exception;
 use MathPHP\LinearAlgebra\Matrix;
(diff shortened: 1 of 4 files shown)


markrogoyski:
This is a nice move in the direction I was thinking about. Sorry that I have not gotten to it myself. Let's change the namespace to the singular `Decomposition` instead of the plural.

Just to give you an idea of what I've been thinking, the next refactorings I would make is to make the LU class (and all decompositions) an instantiable object with a private constructor, and the `decompose` method would be a static factory method. It would return an instance of itself. The constructor would be a simple method that takes in L, U, and P and assigns them to some private attributes. The Matrix would return this LU object instead of an array of Matrixes. The LU object would provide access to each individual Matrix. And as a bonus, the LU object might have a `solve` method to solve a system of linear equations.

Going further, the Matrix would just pull the LU decomposition out of its Catalog, and that would be the lazy loading mechanism.

Everything other than the Matrix class returning the LU object could be done now in Version 1. I don't want to change the interface of what the Matrix returns at this point because it would break how someone might be using it, and would save the public interface breaking changes for Version 2. I'm thinking to get Version 2 out pretty soon after Version 1.

markrogoyski:
BTW, it would probably be good to coordinate refactorings so we both don't work on the same stuff by accident.

Beakerboy:
This literally took 10 minutes to do, and I threw it on here as much as a request for comments as anything. Feel free to reject it in lieu of whatever work you’ve done, or change it any way you see fit. I wouldn’t take it personally.

I have a SVD patch in the wings and I thought I’d line out the thoughts from the issues discussion I posted before submitting that. I figure if we can move in that direction within the v1.0 code-base in a way that existing users wouldn’t see, then I’d do it.

markrogoyski:
I'll merge this and continue doing these refactorings. I enjoy refactoring and moving stuff around. For SVD, do it similar to this as a separate class. We can work out the details later.

markrogoyski:
Nice. Thanks for doing this. Looks good. I'll take it from here and continue this refactor.

PR #338, opened 2019-08-20
SVD

Beakerboy:
* Created SVD class
* Added SVD to MatrixCatalog
* Added SVD to Matrix Functions
* Added isRectangularDiagonal function to Matrix Class
* Added Unit Test


markrogoyski:
Nice. Thanks for working on another important matrix feature!

tests/LinearAlgebra/MatrixDecompositionsTest.php
```diff
@@ -869,4 +869,75 @@ public function dataProviderForQrDecompositionResultingInSquareMatrices(): array
...
+        $A = MatrixFactory::create($A);
+        
+        // When
+        $svd = $A->SVD();
+        
+        // Then
+        $this->assertTrue($svd->U->isOrthogonal());
+        $this->assertTrue($svd->S->isRectangularDiagonal());
+        $this->assertTrue($svd->V->isOrthogonal());
+    }
```
markrogoyski: Here are some other property/axiom tests you can add.

UUᵀ = I
```php
        // Given
        $svd = $A->SVD();
        $I   = MatrixFactory::identity($svd->U->getN());

        // When
        $UUᵀ = $svd->U->multiply($svd->U->transpose());

        // Then
        $this->assertEquals($I->getMatrix(), $UUᵀ->getMatrix());
```

VVᵀ = I
```php
        // Given
        $svd = $A->SVD();
        $I = MatrixFactory::identity($svd->V->getN());

        // When
        $VVᵀ = $svd->V->multiply($svd->V->transpose());

        // Then
        $this->assertEquals($I->getMatrix(), $VVᵀ->getMatrix());
```

src/LinearAlgebra/Decomposition/SVD.php
```diff
@@ -0,0 +1,128 @@
...
+use MathPHP\LinearAlgebra\Matrix;
+use MathPHP\LinearAlgebra\MatrixFactory;
+
+/**
+ * Singular value decomposition
+ *
+ * The generalization of the eigendecomposition of a square matrix to an m x n matrix
+ * https://en.wikipedia.org/wiki/Singular_value_decomposition
+ */
+class SVD extends DecompositionBase
```
markrogoyski: The SVD class doesn't need to implement the `\ArrayAccess` interface like the other decomposition classes. The reason I added it for LU and Crout and Cholesky was because I wanted to make the Matrix return the Decomposition object, but I did not want to break anyone's code who might currently be using the previous return value which is an array. But going forward I think strongly typed objects being returned from the Matrix methods are the way to go, and there is no need to add array compatibility with new objects, as no one is using SVD as an array to have their code broken if it stops working like an array.

It is generally bad practice to break backwards compatibility within point releases. You would wait for a major version release to change the public interface. Semantic versioning actually makes an exception for a pre-1.0 project, but I'd still like to not break people's code as much as possible even pre-1.0.

Anyways, it is fine to extend the base class and provide the extra interface, but it is also not necessary. In version 2.0 I would remove all the legacy array interface for these classes.

tests/LinearAlgebra/MatrixDecompositionsTest.php
```diff
@@ -869,4 +869,75 @@ public function dataProviderForQrDecompositionResultingInSquareMatrices(): array
...
+        $svd = $A->SVD();
+        $svdU = $svd->U;
+        $svdS = $svd->S;
+        $svdV = $svd->V;
+
+        // Then A = USVᵀ
+        $this->assertEquals($A->getMatrix(), $svdU->multiply($svdS)->multiply($svdV->transpose())->getMatrix(), '', 0.00001);
+
+        // And S is expected solution to SVD
+        $this->assertEquals($S->getMatrix(), $svdS->getMatrix(), '', 0.00001);
```
markrogoyski: To make a more robust unit test that will not be brittle and break if the SVD algorithm is refactored, perhaps just check that the diagonal elements of the ∑ result array contain the expect values, regardless of ordering. I'd add matrix unit tests for all the edge cases in this test.

Then, in a separate unit test with the A = U∑Vᵀ and UUᵀ = I and VVᵀ = I tests, you should be good to add a bunch of existing data providers to the unit test and test a wide variety of matrices.

tests/LinearAlgebra/MatrixDecompositionsTest.php
Beakerboy: this is the definition of isOrthoganol()

tests/LinearAlgebra/MatrixDecompositionsTest.php
Beakerboy: R states in their documentation that the values in their SVD function shall be returned ordered from largest to smallest. We could do the same if we wanted.

src/LinearAlgebra/Decomposition/SVD.php
Beakerboy: Since so much of that code was shared, I thought I would consolidate it in a base class...and since I was doing that, making all decompositions consistent in the pre-1.0 code-base makes sense to me.

tests/LinearAlgebra/MatrixDecompositionsTest.php
markrogoyski: Yeah, you are right. My mistake.

src/LinearAlgebra/Decomposition/SVD.php
markrogoyski: Yeah no worries. Making all of them the same right now probably makes the most sense.

tests/LinearAlgebra/MatrixDecompositionsTest.php
markrogoyski: OK. How about doing that. Then it is understood how to write unit tests, and it allows code to be refactored and still be verified by the same tests.

src/LinearAlgebra/Decomposition/SVD.php
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Get S
+     *
+     * @return Matrix
+     */
+    public function getS(): Matrix
+    {
+        return $this->S;
+    }
+
```
markrogoyski: How about adding a method `getSingularValues` that returns the diagonal elements of the matrix Σ (S) of the decomposition. It would return an array of singular values.

Beakerboy: I was planning on adding $this->D if that’s OK...should it be an array or a Vector?

markrogoyski: OK. That is how R does it. Vector works. You can easily turn that into an array if you wanted to.

markrogoyski:
I found some test cases that aren't matching R/Python NumPy. Specifically, one of the singular values seems to have the wrong sign in each case.

Example 1
```php
[
    [
        [8, -6, 2],
        [-6, 7, -4],
        [2, -4, -3],
    ],
    [ 'S' => [
            [14.528807, 0, 0],
            [0, -4.404176, 0],
            [0, 0, 1.875369],
        ],
    ]
],
```
It came up with this for S
```
[14.528806882214, 2.2204460492503E-15, 9.9920072216264E-16]
[1.8873791418628E-15, -4.4041756028697, 1.2878587085652E-14]
[9.1593399531575E-16, 1.2934098236883E-14, 1.8753687206556]
```

R and Numpy have the 4.404 value as being positive.
```R
> a
     [,1] [,2] [,3]
[1,]    8   -6    2
[2,]   -6    7   -4
[3,]    2   -4   -3
> svd(a)
$d
[1] 14.528807  4.404176  1.875369
```
```python
>>> a = array([[8, -6, 2], [-6, 7, -4], [2, -4, -3]])
>>> print(a)
[[ 8 -6  2]
 [-6  7 -4]
 [ 2 -4 -3]]
>>> U, s, VT = svd(a)
>>> print(s)
[14.52880688  4.4041756   1.87536872]
```

Example 2
```php
            [
                [
                    [1, 2],
                    [3, 4],
                    [5, 6],
                ],
                [ 'S' => [
                        [9.52551809, 0],
                        [0, 0.51430058],
                    ],
                ]
            ],
```
It came up with this for S
```php
[9.5255180915651, 8.8817841970013E-16]
[5.5511151231258E-17, -0.51430058065864]
[1.3940498563602E-15, 9.6495784403835E-15]
```

R and Numpy have the 0.5143006 value being postive.
```R
> m
     [,1] [,2]
[1,]    1    2
[2,]    3    4
[3,]    5    6
> svd(m)
$d
[1] 9.5255181 0.5143006
```
```Python
>>> m = array([[1, 2], [3, 4], [5, 6]])
>>> print(m)
[[1 2]
 [3 4]
 [5 6]]
>>> U, s, VT = svd(m)
>>> print(s)
[9.52551809 0.51430058]
```

Beakerboy:
My understanding is that singular values are supposed to always be positive, I’ll look into this.

Beakerboy:
Multiplying U (or V) by a signature matrix will produce an equally valid set of orthogonal eigenvectors from MMᵀ or MᵀM, but the arithmetic will give different signs to S. It looks like I have to create a signature matrix from S and set U = U * Sig.

Beakerboy:
I fixed the negative values by using a signature matrix.

src/LinearAlgebra/Decomposition/SVD.php
Beakerboy: Done

src/LinearAlgebra/Decomposition/SVD.php
```diff
@@ -0,0 +1,157 @@
...
+     *
+     * @throws Exception\MatrixException
+     */
+    public function __get(string $name): Matrix
+    {
+        switch ($name) {
+            case 'U':
+            case 'S':
+            case 'V':
+            case 'D':
```
markrogoyski: Since D is a Vector this will break with the current return type of Matrix.

Beakerboy: yes it would....

Beakerboy: fixed

markrogoyski:
I found another edge case. A 1x1 matrix fails. It looks like it is because the Jacobi eigenvalue method requires a 2x2 matrix or higher. Maybe there needs to be some conditional logic to special handle a 1x1 matrix?

For reference, here is how R and Python handle a 1x1 matrix SVD.
```R
> O <- rbind(c(3))
> O
     [,1]
[1,]    3
> svd(O)
$d
[1] 3

$u
     [,1]
[1,]    1

$v
     [,1]
[1,]    1
```

```python
>>> m = array([[3]])
>>> print(m)
[[3]]
>>> U, s, VT = svd(m)
>>> print(s)
[3.]
```

Beakerboy:
I already have a fix for this in the pull queue. I made the eigenvalue method "smart". instead of specifying jacobiMethod in SVD we can just call eigenvalues() and let the eigenvalue method figure out which algorithm is best.

markrogoyski:
Nice. It sounds like the Smart Eigenvalue PR should be merged first, and then that 1x1 unit test can be added here and it will pass?

Beakerboy:
yes.

Beakerboy:
1x1 edge case passes.

Beakerboy:
I added a little more coverage to get it to a green check mark. Pull anytime.

markrogoyski:
Thanks for continuing to improve this. I'll experiment with it a little more and merge it, presuming I don't find any other edge cases. Thanks.

markrogoyski:
Here is a test case that is failing and seems to differ from R/Numpy.

```php
                [
                    [1, 1, 1, 1, 1],
                    [1, 1, 1, 1, 1],
                    [1, 1, 1, 1, 1],
                ],
                [
                    'S' => [
                        [3.872983, 0, 0, 0, 0],
                        [0, 1.812987e-16, 0, 0, 0],
                        [0, 0, 1.509615e-32, 0, 0],
                    ],
                ],
```

This test is failing
```php
$this->assertTrue($svd->getV()->isOrthogonal());
```

Results from function
```
S
[3.8729833462074, -4.4408920985006E-16, 0, 0, -2.2204460492503E-16]
[-4.9650683064945E-16, 4.9303806576313E-32, 0, 0, 0]
[2.4825341532473E-16, -2.4651903288157E-32, 0, 0, 0]
U
[0.57735026918963, -0.40824829046386, 0.70710678118655]
[0.57735026918963, -0.40824829046386, -0.70710678118655]
[0.57735026918963, 0.81649658092773, 0]
V
[0.44721359549996, 0.22360679774998, 0.22360679774998, 0.22360679774998, 0.28867513459481]
[0.44721359549996, 0.22360679774998, 0.22360679774998, 0.22360679774998, 0.28867513459481]
[0.44721359549996, 0.22360679774998, 0.22360679774998, 0.22360679774998, 0.28867513459481]
[0.44721359549996, 0.22360679774998, 0.22360679774998, 0.22360679774998, -0.86602540378444]
[0.44721359549996, -0.89442719099992, -0.89442719099992, -0.89442719099992, 0]
```

R
```R
> a = rbind(c(1,1,1,1,1),c(1,1,1,1,1),c(1,1,1,1,1))
> a
     [,1] [,2] [,3] [,4] [,5]
[1,]    1    1    1    1    1
[2,]    1    1    1    1    1
[3,]    1    1    1    1    1
> svd(a)
$d
[1] 3.872983e+00 1.812987e-16 1.509615e-32

$u
           [,1]       [,2]          [,3]
[1,] -0.5773503  0.8164966 -1.751211e-16
[2,] -0.5773503 -0.4082483 -7.071068e-01
[3,] -0.5773503 -0.4082483  7.071068e-01

$v
           [,1]       [,2]          [,3]
[1,] -0.4472136  0.8944272  8.316700e-17
[2,] -0.4472136 -0.2236068 -8.660254e-01
[3,] -0.4472136 -0.2236068  2.886751e-01
[4,] -0.4472136 -0.2236068  2.886751e-01
[5,] -0.4472136 -0.2236068  2.886751e-01
```

Python Numpy
```python
>>> m = array([[1,1,1,1,1], [1,1,1,1,1], [1,1,1,1,1]])
>>> print(m)
[[1 1 1 1 1]
 [1 1 1 1 1]
 [1 1 1 1 1]]
>>> U, s, VT = svd(m)
>>> print(s)
[3.87298335e+00 1.81298661e-16 1.50961461e-32]
>>> print(U)
[[-5.77350269e-01  8.16496581e-01 -1.75121059e-16]
 [-5.77350269e-01 -4.08248290e-01 -7.07106781e-01]
 [-5.77350269e-01 -4.08248290e-01  7.07106781e-01]]
>>> print(VT)
[[-4.47213595e-01 -4.47213595e-01 -4.47213595e-01 -4.47213595e-01
  -4.47213595e-01]
 [ 8.94427191e-01 -2.23606798e-01 -2.23606798e-01 -2.23606798e-01
  -2.23606798e-01]
 [ 8.31669992e-17 -8.66025404e-01  2.88675135e-01  2.88675135e-01
   2.88675135e-01]
 [ 2.00841232e-17 -3.54270280e-17 -5.77350269e-01  7.88675135e-01
  -2.11324865e-01]
 [ 0.00000000e+00  0.00000000e+00 -5.77350269e-01 -2.11324865e-01
   7.88675135e-01]]
```

Beakerboy:
Interesting, that means the Jacobi method is not finding the correct vectors on a 5x5 matrix of all ones. 

Have you noticed that R has different matrix sizes than we do? Do you tho k we should have a flag to either return a square S matrix or a rectangular S?

Beakerboy:
No...it’s the Find eigenvectors from eigenvalues function. It must not like something about all the duplicate eigenvalues. The reduced row echelon form on a ones matrix might be presenting challenges.

markrogoyski:
Are you sure there is an issue with RREF? I went through that test case in the eigenvectors method and dumped every matrix I thought it was computing the RREF of and created a unit test for each of them just now (in develop branch: https://github.com/markrogoyski/math-php/commit/5a83e63e4239760f550c004ac15e8c0e11350d8e). It seemed to compute the correct RREF for each matrix. Let me know if I missed a matrix and I can write some other unit tests for RREF to validate that it is working correctly.

Beakerboy:
The “findEigenvectors” function operates off the source matrix in RREF form, so since this matrix will be weird in rref, it might give weird vectors.

markrogoyski:
Any updates on this?

Beakerboy:
You keep finding edgier edge cases...😄 I’ll keep working on it.

Beakerboy:
I'm closing this and resubmitting a newer version.

PR #343, opened 2019-08-30
Find roots of any polynomial.

Beakerboy:


src/LinearAlgebra/Decomposition/Eigen.php
```diff
@@ -0,0 +1,105 @@
...
+    public static function powerIteration(Matrix $A, int $iterations = 100000): array
+    {
+        if (!$A->isSquare()) {
+            throw new Exception\BadDataException('Matrix must be square');
+        }
+        
+        $b    = MatrixFactory::random($A->getM(), 1);
+        $newμ = 0;
+        $μ    = -1;
+        while (!Support::isEqual($μ, $newμ, 1e-15)) {
```
markrogoyski: I don't know if it is the algorithm or not, but the 1e-15 seemed to be too precise on my machine. On my machine the unit tests would fail about 1 out of 3 times. The pull request Travis build also showed this inconsistent behavior.

Beakerboy: You’re probably right. I set it low while troubleshooting it and never increased after it started working and ran it a few times without issue. I can add more tests and tweak stuff. I have not tested how it will fail with imaginary roots.

markrogoyski:
Any update on this?

Beakerboy:
Nothing groundbreaking. Release without it.

Beakerboy:
I'll lessen the scope of this in a new pull request. instead of trying to use the companion matrix to calculate root off the bat. I'll just return the matrix.

PR #344, opened 2019-09-24
Arbitrary Length Integer Preview

Beakerboy:
Here's a preview of an Arbitrary Length Integer class. Look it over and ask questions.


tests/Number/ArbitraryIntegerTest.php
```diff
@@ -0,0 +1,303 @@
...
...60176999612831860788386150279465955131156552036093988180612138558600301435694527224206344631797460594682573103790084024432438465657245014402821885252470935190620929023136493273497565513958720559654228749774011413346962715422845862377387538230483865688976461927383814900140767310446640259899490222221765904339901886018566526485061799702356193897017860040811889729918311021171229845901641921068884387121855646124960798722908519296819372388642614839657382291123125024186649353143970137428531926649875337218940694281434118520158014123344828015051399694290153483077644569099073152433278288269864602789864321139083506217095002597389863554277196742822248757586765752344220207573630569498825087968928162753848863396909959826280956121450994871701244516461260379029309120889086942028510640182154399457156805941872748998094254742173582401063677404595741785160829230135358081840096996372524230560855903700624271243416909004153690105933983835777939410970027753472000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'
```
markrogoyski: That's impressive. With the regular iterative algorithm and PHP number types you will get a result of infinity after factorial of 300 or so.

markrogoyski:
Thanks for the pull request preview. I'll check it out.

tests/Number/ArbitraryIntegerTest.php
Beakerboy: In theory, this object should be able to calculate 1250000000000000000! before overflowing. If all functions that call `strlen()` are altered, so we beaver have to measure the length, the only limitation would be memory and page file size.

Beakerboy: Also remember, php overflows with 300! As a float, and this gives the precise Integer value.

Beakerboy: The maxint that this object can save is on the order of 10^(1E50) or 10^10^50.

markrogoyski:
Thanks for the pull request. I'm still learning more about this.

My initial thought is that maybe there are two classes here? A BaseEncoderDecoder utility class that the ArbitraryInteger class could use that would simplify ArbitraryInteger a bit and allow it to focus on numeric responsibilities and not so much on alphabet/base conversions. Would that distinction of responsibilities be helpful here?

Beakerboy:
The utility class isn’t a bad idea. I’ll have to think about how to implement it. As far as I see, this is the only class that would use it, so maybe a compromise would be to separate some of the logic into protected static functions. Some of the base256 arithmetic could be better served this way.

Beakerboy:
Let me know if there are any function or features you feel should be included. I was thinking that `ArbitraryInteger::rand()` is one, maybe `isPrime()`. I’m also trying to figure out the best way to handle negative numbers. This is a good motivation for moving the “non-standard alphabet” stuff out of this class.

markrogoyski:
I'd say two advantages of an external utility class, even if it is only used by a single class, is that you can easily test the specific functionality, gaining confidence that it works, and the ArbitraryInteger class is easier to understand. Alphabet encoding/decoding is just a detail that "works." The ArbitraryInteger class can focus on interesting numeric stuff.

Beakerboy:
I don't know if you noticed, but I have `fact()` and `ackerman()` as static functions. The more I think about it, `fact()` should probably be nono-static, since it is an operator acting on a number. However, Ackerman is more of a function of two numbers, so maybe it should stay static. What do you think? I mainly have it in there as a functional test, not because anyone actually uses the Ackerman Function as far as I know. That being the case, I ~could~ just remove it. Let me know what you think.

markrogoyski:
In PHP, regular ints are not objects and you can't call methods on them, so to make things consistent with how things like factorial work on regular ints, it would seem reasonable to have an ArbitraryInteger[Something] class that performed operations on ArbitraryIntegers. However, they are already objects and could just call methods.

Maybe it is a matter of scope. If you intend to create all sorts of statistical and combinatorics and other functions to work on these ArbitraryIntegers, then it probably makes sense to have the appropriately named classes with static methods to do them. But if the idea is just to have a couple to show what these guys can do, then having `fact` on the object itself is fine and probably preferable.

For `ackermann`, I think it depends no how you answer the previous question. It seems more natural as a method existing elsewhere. But if you are thinking these are all the specific ArbitraryInteger methods that are going to exist, then a single static method on ArbitraryInteger that works on ArbitraryIntegers seems fine to me. It also showcases what the class can do.

markrogoyski:
Maybe a better model is the Complex class. That represents a number and has lots of methods on it. It already exists and set a pattern for number-like objects. I'd say leave fact as an object method and keep adding methods on the object unless it seems like it is getting out of control, then it can reconsidered at that time.

Beakerboy:
The GMP extension would probably be a good resource too, since it is already a well known arbitrary precision integer resource for PHP

Beakerboy:
I moved out the exotic number base handling, but I’m not really happy with it, yet don’t know what would be better. I also added support for negative integers. I just realized that negative multiplication and division are not in there yet.

src/Functions/BaseEncoderDecoder.php
```diff
@@ -0,0 +1,136 @@
...
+                    $chr = $number[$i];
+                    $number[$i] = chr(strpos($offset, $chr));
+                }
+            }
+        }
+        // Convert to base 256
+        $base256 = new ArbitraryInteger(0);
+        if ($base <= 256) {
+            $base_obj = new ArbitraryInteger($base);
+            $place_value = new ArbitraryInteger(1);
```
markrogoyski: What is the purpose of these two variables? Did you intend to use them for something?

Beakerboy: I think those were used some time ago, but it led to excessive recursion, when creating new objects within the constructor.

Beakerboy: This has been fixed

Beakerboy:
I believe everything is ready to go on this pull request.

markrogoyski:
Thanks for continuing to work on this. I'll try to review this soon. I've been quite busy with a lot of things.

Beakerboy:
No problem. The one thing I was thinking would be advantageous would be to have a IntegerInterface, which would be used internally in the Aackerman function for example. This way I can develop a MathPhpGMP object externally, and allow users to leverage the huge performance gains from GMP, while still using the advanced objects and algorithms in MathPHP. The structure of the interface would extend ObjectArithmatic some to include the most important functions in the [GMP](https://www.php.net/manual/en/book.gmp.php) library

Beakerboy:
https://github.com/Beakerboy/math-php/blob/IntegerInterface/src/Functions/ArbitraryIntegerFunctions.php
https://github.com/Beakerboy/math-php-gmp

Beakerboy:
Alternatively, while GMP would not be a requirement, MathPHP could detect if it is present and use GMP functions when available.

markrogoyski:
Yeah. Using something if it is available is fine, as long as it is not a requirement, and there is a fallback mechanism for when it is not there that works in an equivalent way.

markrogoyski:
One advantage of a separate project/library is that you can be explicit in the requirements (composer.json) that it requires the ext-gmp extension. You know upfront what you need to get it to work. This doesn't help a more casual user or even a power user who just isn't familiar with everything involved with PHP extensions and arbitrary length numbers.

On the other hand, an advantage of having MathPHP be smart and do the best thing based on what's available is that the user doesn't need to know anything and it just works, and maybe works well with the GMP extension, but maybe not.

You could even do both.

Interfaces are good. I'm curious, is your proposed interface something that would be exposed externally? What are you thinking the use case is?

Beakerboy:
I have links above to my “Interfaced” version of the ArbitraryIntegerFunctions class as well as my separate GMPInteger object. My library contains a new object that implements the same interface. This new object is able to run the ArbitraryInteger unit test suite in less than one minute. The one very long commented-out Ackerman case takes several minutes to run on Travis without GMP.

markrogoyski:
Can you try merging the latest develop into your ArbitraryInteger branch? I'm guessing you haven't done this in awhile because it is showing 309 files changed and my phone can't load the diff!

markrogoyski:
For the `Integer` interface, I was thinking the `public function toInteger(): int` method might be clearer if it was called `toInt` or even `toScalarInt`, as `Integer::toInteger` seems like a no-op just from the name.

Beakerboy:
I think the diff you see are all the changes from when the fork was made. I merged develop back into ArbitraryInteger some time ago. I did it again and the "files changed" increased.

If you would prefer, I can make a clean pull request straight off of develop that only has the new files as they are now.

markrogoyski:
It's strange. I'd expect github to resolve all the other changes as already implemented and just pick up the actual new changes. I wonder if github is doing the diff based on the state of the develop branch when you originally opened the pull request vs. what the state is currently.

I'll just pull from your branch and take a look at the code one more time. Is it all ready to go now?

Beakerboy:
I just pushed changing the name of toInt(). It should be good now.

markrogoyski:
Sorry for the delay in reviewing this. A couple minor naming and location issues.

`getPositive` returns a float, so it is more natural to word these kinds of methods as `isPositive`.

`getBinary`, `toFloat`, `toInt`: Let's try to get the language consistent. I realize `getBinary` isn't doing any actual work, but that is an implementation detail. Maybe word them all as `to(Binary|Float|Int)`, or consider `as(Binary|Float|Int)`.

The `ArbitraryIntegerFunctions` class is already in a namespace `Functions` so the Functions suffix is redundant. Namespaces allow us to have classes with the same names, so I think it is OK to have `Number\ArbitraryInteger` and `Functions\ArbitraryInteger`.

`fact` seems like it belongs in `Functions\ArbitraryInteger`.

The `rand` method is an interesting one. Is this a factory method or a utility? Maybe this would be more intuitive to go in the `Functions\ArbitraryInteger` integer utility class? What do you think?

Beakerboy:
I was conflicted on whether fact should be static or not. Is there an advantage to `$fact = $number->fact()` versus `$fact = ArbitraryInteger::fact($number)`. I’m actually leaning towards leaving it in the class and removing the static designation.

I think `rand()` should go the the functions class.

markrogoyski:
An important concept in object-oriented programming is cohesion. Do the properties and methods in a class all belong together. In a class with private data properties, you usually answer this question by looking if the methods on the class work on the same data in the class. If you have properties A B and C and method 1 works with A, method 2 works with B, and method 3 works with C, then the class is not very cohesive.

A static method on an instance of a class will never access its internal data, so it detracts from the overall cohesiveness of the class. That's not to say it never makes sense to have static methods on an object instance, but it requires a unifying theme to make them all work.

In this case, `fact` isn't related to anything in a particular instance of ArbitraryInteger, so it doesn't need to be an instance method. If you just want to know the factorial of a number, why would you need to construct some arbitrary object (pun intended) just to compute the factorial. Static makes sense here. But since it is static, it raises the question should it even be in this class, or go in some utility class instead.

Some reading material:
https://en.wikipedia.org/wiki/Cohesion_(computer_science)

The other concept that tends to go hand and hand with cohesion is coupling:
https://en.wikipedia.org/wiki/Coupling_(computer_programming)

Ideally you want high cohesion within your class, and loose coupling between your classes.

Beakerboy:
I’ll have to dig into the material, because I don’t see how $number->fact() and $number->sqrt() are really any different. They can both be implemented statically as ArbitraryInteger::sqrt() or ArbitraryInteger::fact(). I guess you could make an argument that factorial is just a series of multiplications, but so is exponentiation, so we should also pull out `pow()`. /devils_advocate

Beakerboy:
Ok, good, I’m not crazy. My internal debate is whether factorial should be static, and I. The functions class, or non-static, and remain in the object class. If non-static, it would operate on $this, and not take a parameter. Alternatively, the existing factorial could be extended to return an ArbitraryInteger if an overflow will occur, or if provided one as input.

Beakerboy:
This should be it!

markrogoyski:
Thanks for working on this and incorporating some feedback suggestions. Looks good!

markrogoyski:
What should ArbitraryInteger `isqrt(0)` evaluate to? I thought 0, but it seems to throw a DivisionByZerrorError when I try it.

PR #348, opened 2019-10-07
Power Iteration-perturb and resolve.

Beakerboy:


Beakerboy:
Well, so much for this idea. My only thoughts are:

- I noticed $b is not a unit vector in the beginning. Scaling this might help.
- I wonder what the iteration count is for the failures? Maybe we need a minimum number of iterations.

markrogoyski:
At one point I tried making it run 1,000 iterations regardless of the stopping condition and I still got random errors, probably due to bad luck with the initial random vector.

Beakerboy:
This seems more promising. Instead of perturbing one element of the eigenvector, I’m perturbing all of them. Some of the routine failures were occurring on eigenvectors with zero entries, and perturbing one was just rescaling the eigenvector.

Beakerboy:
It passes unit testing 😁. Try this version a bunch and see if there are failures.

markrogoyski:
This seems like good progress. I ran the powerIteration tests hundreds of times and did not get any test failures due to incorrect results, but there were some random errors now. They were rare but did happen a few times in my repeated run.

```
1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testPowerIterationViaMatrix with data set #11 (array(array(1, 4), array(4, 1)), array(5.0, -3.0), 5)
MathPHP\Exception\BadParameterException: Parameter λ cannot equal 0

/math-php/src/LinearAlgebra/Matrix.php:1520
/math-php/src/LinearAlgebra/Eigenvalue.php:207
/math-php/src/LinearAlgebra/Matrix.php:3381
/math-php/tests/LinearAlgebra/EigenvalueTest.php:128
```

markrogoyski:
Running the test suite 1,000 times, this error happened 55 times.

```bash
$ ../vendor/bin/phpunit LinearAlgebra/EigenvalueTest.php --filter=power --repeat=1000

...

ERRORS!
Tests: 87000, Assertions: 86945, Errors: 55.
```

markrogoyski:
The good news is no failures due to incorrect results. So, it either gets the right answer, or crashes.

Beakerboy:
This error must happen when several very small values are randomly selected. I picked a check which will redraw the random matrix of this happens.

markrogoyski:
Looks good!

```
$ ../vendor/bin/phpunit LinearAlgebra/EigenvalueTest.php --filter=power --repeat=1000

...

Time: 2.13 minutes, Memory: 6.00 MB

OK (87000 tests, 87000 assertions)
```

I think it is still theoretically possible, but seems extremely unlikely now.

markrogoyski:
I ran it 10,000 times, and it happened once. Maybe this is tolerable. Thoughts?

```bash
$ ../vendor/bin/phpunit LinearAlgebra/EigenvalueTest.php --filter=power --repeat=10000

...

......................................................... 869934 / 870000 ( 99%)
......................................................... 869991 / 870000 ( 99%)
.........                                                 870000 / 870000 (100%)

Time: 20.28 minutes, Memory: 6.00 MB

There was 1 error:

1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testPowerIterationViaMatrix with data set #41 (array(array(4, 0, 2, -2), array(0, 9, -6, 3), array(2, -6, 5, -3), array(-2, 3, -3, 2)), array(15, 5, 0, 0), 15)
MathPHP\Exception\BadParameterException: Parameter λ cannot equal 0

/math-php/src/LinearAlgebra/Matrix.php:1520
/math-php/src/LinearAlgebra/Eigenvalue.php:222
/math-php/src/LinearAlgebra/Matrix.php:3381
/math-php/tests/LinearAlgebra/EigenvalueTest.php:128
```

Beakerboy:
This last error is happening in a different place than the earlier one. Do you know which test this was? Was it some weird matrix with all zero eigenvalues or something?

markrogoyski:
The error message shows the input matrix:

```
1) MathPHP\Tests\LinearAlgebra\EigenvalueTest::testPowerIterationViaMatrix with data set #41 (array(array(4, 0, 2, -2), array(0, 9, -6, 3), array(2, -6, 5, -3), array(-2, 3, -3, 2)), array(15, 5, 0, 0), 15)
MathPHP\Exception\BadParameterException: Parameter λ cannot equal 0
```

So it would be this matrix:
```
[
    [4, 0, 2, -2],
    [0, 9, -6, 3],
    [2, -6, 5, -3],
    [-2, 3, -3, 2],
]
```
with these expected eigenvalues: ```[15, 5, 0, 0]```, and an expected dominant eigenvalue of ```15```.

Beakerboy:
The only way to have a zero frobenius norm is if the eigenvector is zero, and zero is not a valid eigenvector. So if would probably make sense to just re-randomize the b Vector in this case, no?

markrogoyski:
This time 10,000 runs and no errors.

```bash
../vendor/bin/phpunit LinearAlgebra/EigenvalueTest.php --filter=power --repeat=10000

...

Time: 20.53 minutes, Memory: 6.00 MB

OK (870000 tests, 870000 assertions)
```

I'll repeat it a few more times.

Beakerboy:
Statistically if something has a 1in X chance of happening, and you do said thing X number of times, then there is about a 1/3 chance of it not happening.

Beakerboy:
With these new “safety features” to prevent division by zero, coveralls will tell how often they are triggered. If you insist on covering the super rare condition inside the loops, we will have to mock the calls to `rand()` like I did elsewhere. I’d say it could be done by performing a power iteration on the matrix $A:

    [[1, 0],
     [0, 0]]

Using the initial $b of

    [[0],
     [1]]

This would force the value of $A->multiply($b) to be the zero matrix, but when $b is then set to a new random vector, it should find the solution.

markrogoyski:
Thanks for the input suggestion. I'll try writing the edge case unit tests. Thanks for working on this.

Beakerboy:
Similarly, if you want to make sure the first safety check is always activated, a zero matrix will have to be created by the first call to `rand()`. This would ensure that out coverage score does not voice up and down.

PR #349, opened 2019-10-09
Nilpotent Matrix Check

Beakerboy:
A nilpotent Matrix has all zero eigenvalues.


markrogoyski:
Neat. Never heard of this.

src/LinearAlgebra/Matrix.php
```diff
@@ -827,6 +827,36 @@ public function isIdempotent(): bool
...
+     */
+    public function isNilpotent(): bool
+    {
+        if ($this->trace() !== 0) {
+            return false;
+        }
+        $A = $this;
+        $m = $this->getM();
+        $zero = MatrixFactory::zero($m, $m);
+        for ($i = 1; $i < $m; $i++) {
```
markrogoyski: If a matrix is nilpotent, is it guaranteed to turn all zeros by A^m? The definition on Wikipedia seemed to suggest that there is some power you can raise it to that will be all zeros, but I didn't read it as saying it would be by m or less powers. However, all the examples on Wikipedia occur at m or less, so maybe it is a safe assumption?

Beakerboy: I had heard about it until today. I spent a little time on using power iteration to find all eigenvalues using dimensionality reduction. It worked on all the matrixes in the main dataProvider except for one with a zero. It throws an exception on too many iterations. When the eigenvalue is zero, the Vector will be zero, but the current code redraws the random $b vector when it is zero. I ~could~ check the determinant to see if a matrix has at least one zero in the beginning, but that does not help if there are more than one. Instead, I can check if the matrix is nilpotent after each solution is found. If it is, I can set all the remaining eigenvalues to zero.

Beakerboy: Wikipedia says, "The degree of an n×n matrix is always less than or equal to n. For example, every 2×2 nilpotent matrix squares to zero.

Beakerboy: Also, there is one line not covered. I’m thinking it is possible for there to be a matrix with a trace of zero, which does not have a trace of zero after being raised to a power less than n. Since a nilpotent Matrix should always have a zero trace, I figured I would check that, and return early if possible.

markrogoyski: OK thanks. I didn't notice that in the _Characterization_ section.

markrogoyski:
Looks good. Let me know if you are done adding things and I'll merge it.

Beakerboy:
I have one more test case.

src/LinearAlgebra/Matrix.php
```diff
@@ -827,6 +827,36 @@ public function isIdempotent(): bool
...
+     * A square MxM matrix is nilpotent if it becomes the
+     * zero matrix when raised to some power k<=M.
+     *
+     * NilPotent matricies will have a zero trace for all k
+     *
+     * https://en.wikipedia.org/wiki/Nilpotent_matrix
+     */
+    public function isNilpotent(): bool
+    {
+        if ($this->trace() !== 0) {
```
markrogoyski: Should you also check that it is square first before continuing?

Beakerboy: I figure since `trace()` already checks that then there’s no need, no?

markrogoyski: Trace throws an exception
```php
        if (!$this->isSquare()) {
            throw new Exception\MatrixException('trace only works on a square matrix');
        }
```

`isNilpotent` could just return false instead of blowing up.

Beakerboy: I realized how stupid my thought was just a minute ago.

markrogoyski:
I think the matrix [0] should be nilpotent, but I think based on the loop criteria it will always return false.

Beakerboy:
I could change the loop to a do-while, or check if the matrix is a zero matrix.

src/LinearAlgebra/Matrix.php
```diff
@@ -823,10 +823,46 @@ public function isRref(): bool
...
+        $A = $this;
+        $m = $this->getM();
+        $zero = MatrixFactory::zero($m, $m);
+        for ($i = 1; $i < $m; $i++) {
+            $A = $A->multiply($this);
+            if ($A->isEqual($zero)) {
+                return true;
+            }
+            if ($A->trace() !== 0) {
+                return false;
```
markrogoyski: I like the idea of a return early, but with this inside the loop, I'm thinking maybe the default `return false` at the bottom may be impossible to reach.

Beakerboy: It is impossible for a 2x2, which is easy to prove algebraically, but my guess is a special 3x3 of larger might.

Beakerboy: It looks like any 3x3 matrix with a zero trace when k is 1 and 2 will either be the zero matrix or the identity matrix when k=3. Excel is only finding the identity matrix when generating a random initial matrix of the form:

    A    B    C
    D    E    F   Where  I = -1*(A+E) and D = (((A²+E²+I²)/-2)-G*C-F*H)/B
    G    H    I

Beakerboy: This would be a good candidate for this XKCD:
https://www.xkcd.com/2200/

markrogoyski: I tried stuff like this to see if I could hit that line of code and it never hit.
```php
        foreach (range(1, 10) as $i) {
            foreach (range(1, 1000000) as $_) {
                $A = MatrixFactory::random($i, $i, -20, 20);
                $A->isNilpotent();
            }
        }
```

I've written my share of theoretically unreachable code.... just in case you think...

Beakerboy: The Wikipedia article on “matrix trace” says that all nilpotent matrixes have zero trace for all k, and the corollary is also true, if it has a zero trace for all k, it’s nilpotent. Again, there is nothing that limits it to k<=n, but I think we can assume that.

If “coverage percent” is a metric that is important to optimize, we can be sneaky with the algorithm to optimize for coverage. For example, instead of returning early, we can trigger a break and have a single return statement at the bottom of the function.

markrogoyski: Code coverage for the same of coverage is not the goal, but 100% and beyond is expected, because we should understand what every line of code is for and why it is there, and what scenarios we expect to work.

It sounds like we think there is a scenario where a Matrix is not nilpotent, but will have a 0 trace for every iteration of raising it to a power. Does a matrix exist that would satisfy that scenario? If the answer is "unknown," then we have a "just-in-case" line of code that isn't easily testable and coverable. Although there might be some creative ways to hit it in a unit test but I wouldn't waste too much time on it. It will always be there in the code coverage report to remind you and it would be a low priority to try to write that test.

Beakerboy:
If we want to reach for the unreachable:

     public function isNilpotent(): bool
    {
       if(!$this->isSquare() || $this->trace() !== 0) {
            return false;
        }
        $A = $this;
        $m = $this->getM();
        $zero = MatrixFactory::zero($m, $m);
        if ($this->isEqual($zero)) {
                return true;
        }
        $nilpotent = false;
        for ($i = 1; $i < $m; $i++) {
            $A = $A->multiply($this);
            if ($A->isEqual($zero)) {
                $nilpotent = true;
                break;
            }
            if ($A->trace() !== 0) {
               break;
            }
        }
        return $nilpotent;
    }

markrogoyski:
> If we want to reach for the unreachable:
> 
> ```
>  public function isNilpotent(): bool
> {
>    if(!$this->isSquare() || $this->trace() !== 0) {
>         return false;
>     }
>     $m = $this->getM();
>     $zero = MatrixFactory::zero($m, $m);
>     if ($this->isEqual($zero)) {
>             return true;
>     }
>     $A = $this;
>     $nilpotent = false;
>     for ($i = 1; $i < $m; $i++) {
>         $A = $A->multiply($this);
>         if ($A->isEqual($zero)) {
>             $nilpotent = true;
>             break;
>         }
>         if ($A->trace() !== 0) {
>            break;
>         }
>     }
>     return $nilpotent;
> }
> ```

Rather than rewrite the method, you could just use some mock trickery that will hit the line.
```php
    public function testIsNilpotentMock()
    {
        // Given
        $mockMatrix = $this->getMockBuilder(Matrix::class)
            ->setConstructorArgs([[[1, 2],[3, 4]]])
            ->enableOriginalConstructor()
            ->setMethods(['getM', 'isSquare', 'trace', 'multiply'])
            ->getMock();

        // And
        $mockMatrix->expects($this->any())->method('getM')->willReturn(5);
        $mockMatrix->expects($this->any())->method('isSquare')->willReturn(true);
        $mockMatrix->expects($this->any())->method('trace')->willReturn(0);
        $mockMatrix->expects($this->any())->method('multiply')->willReturn($mockMatrix);

        // When
        $isNilpotent = $mockMatrix->isNilpotent();

        // Then
        $this->assertFalse($isNilpotent);
    }
```

Beakerboy:
This is a decision I’ll leave up to you. I don’t like break statements, but that might be just a personal preference...there’s obviously a time and place for them, and this might be that place. Similarly, I don’t like the idea of having to mock a strange condition in what should be a simple mathematical function. If math theory says that specific condition should not exist, we should have an exception, not a false return. My vote is for break statements.

markrogoyski:
Sure. Let's go with the break. I like understanding how to write tests to hit lines of codes in real scenarios with real data.

Beakerboy:
Done and ready to go!

markrogoyski:
Thanks for the pull request and iterating on it.

PR #351, opened 2019-10-10
DecompositionBase

Beakerboy:
Collecting some of the common functions together in one place. Reducing lines of code, increasing coverage.

### DIFF
--- a/src/LinearAlgebra/Decomposition/Decomposition.php
+++ b/src/LinearAlgebra/Decomposition/Decomposition.php
@@ -14,6 +14,13 @@ abstract class Decomposition implements \ArrayAccess
      **************************************************************************/
 
     /**
+     * @param  mixed $i
+     * @param mixed $i
+     * @return bool
+     */
+    abstract public function offsetExists($i): bool;
+
+    /**
      * @param mixed $i
      * @return mixed
      */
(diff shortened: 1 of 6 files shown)


src/LinearAlgebra/Decomposition/DecompositionBase.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace MathPHP\LinearAlgebra\Decomposition;
+
+use MathPHP\Exception;
+use MathPHP\LinearAlgebra\Matrix;
+
+abstract class DecompositionBase implements \ArrayAccess
```
markrogoyski: How about just naming it `Decomposition` without the Base suffix?

Beakerboy:
Done. Let me know if there’s anything else.

src/LinearAlgebra/Decomposition/Decomposition.php
```diff
@@ -0,0 +1,43 @@
...
+
+use MathPHP\Exception;
+use MathPHP\LinearAlgebra\Matrix;
+
+abstract class Decomposition implements \ArrayAccess
+{
+    abstract public static function decompose(Matrix $M);
+
+    /**************************************************************************
+     * ArrayAccess INTERFACE
```
markrogoyski: How about also adding `offsetExists` as an abstract empty method that each Decomposition has to implement.

Beakerboy: Good idea

markrogoyski:
Looks good. Thanks for continuing to improve the Matrix classes.

PR #353, opened 2019-10-15
Complex Properties

Beakerboy:
Since real numbers are a subset of the complex numbers, complex properties would be appropriate.
* Added conjugateTranspose()
* Added isUnitary()
* Added isHermitian()

### DIFF
--- a/src/LinearAlgebra/Matrix.php
+++ b/src/LinearAlgebra/Matrix.php
@@ -1173,7 +1173,7 @@ class Matrix implements \ArrayAccess, \JsonSerializable
         $AAᴴ = $this->multiply($Aᴴ);
         $AᴴA = $Aᴴ->multiply($this);
 
-        $I = Matrixfactory::identity($A->getM());
+        $I = Matrixfactory::identity($this->m);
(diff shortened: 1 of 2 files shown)


src/LinearAlgebra/Matrix.php
```diff
@@ -1757,6 +1807,18 @@ public function transpose()
...
+    /**
+     * Conjugate Transpose
+     *
+     * Returns the complex conjugate of the transpose. For a real matrix, this is the same as the transpose.
+     *
+     * https://en.wikipedia.org/wiki/Conjugate_transpose
+     */
+    public function conjugateTranspose()
+    {
+        return $this->transpose();
```
markrogoyski: Are you planning on adding something to ObjectMatrix that checks if it is full of Complex numbers and then conjugate each element if they are complex?

Beakerboy: Yes, but I would extend the ObjectMatrix to a ComplexMatrix.

markrogoyski:
Thanks for the pull request.

PR #354, opened 2019-10-24
Factorial optimization

aymeric451:
1. Multiply by 1 is useless
2. Pre-incrementation is faster : https://syframework.alwaysdata.net/pre-increment

PR #355, opened 2019-10-24
Factorial optimization

aymeric451:
Recreated the #354 PR on the develop branch.

1. Multiply by 1 is useless
2. Pre-incrementation is faster : https://syframework.alwaysdata.net/pre-increment

### DIFF
--- a/src/Probability/Combinatorics.php
+++ b/src/Probability/Combinatorics.php
@@ -51,9 +51,9 @@ class Combinatorics
             throw new Exception\OutOfBoundsException('Cannot compute factorial of a negative number.');
         }
         $factorial = 1;
-        while ($n > 0) {
+        while ($n > 1) {
             $factorial *= $n;
-            $n--;
+            --$n;
         }
         return $factorial;
     }


Beakerboy:
Why did you close one and re-open it? This looks like the exact same changes.

aymeric451:
I just changed the target branch to develop.
Other changes are needed ?

Beakerboy:
I didn’t notice the target change. Excellent! I make that mistake all the time. As simple of a change as this is, I don’t see anything else that would need to be in place. Feel free to submit any more improvements or additions that you see fit.

markrogoyski:
Since PHP is implemented in C, and these simple arithmetic operations directly map to C operations, this phenomenon is most likely because of how C treats pre- and post-increment operations.

I found this documentation that adds details of why this is the case in C:
https://softwareengineering.stackexchange.com/questions/358725/pre-increment-vs-post-increment
https://stackoverflow.com/questions/2020184/preincrement-faster-than-postincrement-in-c-true-if-yes-why-is-it
https://www.includehelp.com/tips/c/pre-increment-is-faster-than-post-increment.aspx

The basic idea is that the post operations keep a copy of `i` whereas the pre operators don't.

You would think that the complier would optimize this but I guess not in certain cases.

markrogoyski:
Thanks for the pull request and optimization improvement.

PR #359, opened 2019-11-27
optimize prime factorizations

Theriault:
Not the best algo, but an improvement

prime factor 2..1e5:
before: 15.95s
after: 0.09s

### DIFF
--- a/src/NumberTheory/Integer.php
+++ b/src/NumberTheory/Integer.php
@@ -120,14 +120,34 @@ class Integer
 
         $remainder = $n;
         $factors   = [];
-        $divisor   = 2;
 
-        while ($remainder > 1) {
+        foreach ([2, 3] as $divisor) {
             while ($remainder % $divisor === 0) {
                 $factors[] = $divisor;
                 $remainder = intdiv($remainder, $divisor);
             }
-            $divisor++;
+        }
+
+        $divisor = 5;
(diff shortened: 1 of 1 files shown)


markrogoyski:
Hi,
Thanks for your interest in MathPHP and improving it.

The improvement looks impressive! Can you document, or just reply here with what the name of the algorithm is or a link to something that describes it so I can better understand it while reviewing the code.

Thanks again!
Mark

Beakerboy:
I think I get it. Instead of increment if the divisor by one, it’s only looking for odd factors `$i +=2` but skipping every third with`$i +=4` Because they are divisible by 3. Instead of going all the way to the dividend, it stops at the square root because the smallest factor of any number must be less than or equal to the square root of a number.

Theriault:
Beakerboy explained it well. The algo is:
1. let `n` be the ongoing remainder
2. try prime factoring `n` with 2 and 3
3. try prime factoring `n` with increasing ℕ of the form `6𝑘±1` up through `√n`  (all other ℕ are divisible by 2 and/or 3)
4.  if `n` is still > 1, the remainder is a prime factor

markrogoyski:
Thanks for the pull request and the nice improvement of the prime factorization algorithm. This will be in the next release.

Thanks again!

PR #360, opened 2019-12-01
Number theory additions

Theriault:
This adds the following functions:
- `Integer::numberOfDivisors($n)`
- `Integer::sumOfDivisors($n)`
- `Integer::aliquotSum($n)`
- `Integer::isAbundantNumber($n)`
- `Integer::isDeficientNumber($n)`
- `Integer::radical($n)`
- `Integer::isSquarefreeInteger($n)`
- `Integer::totient($n, $k = 1)`
- `Integer::cototient($n)`
- `Integer::reducedTotient($n)`
- `Integer::mobius($n)`
- `Integer::isRefactorableNumber($n)`
- `Integer::isSphenicNumber($n)`

It refactors `Integer::isPerfectNumber($n)` to work off the prime factorization. With the optimization for prime factorization applied, this reduces testing 2...1e6 from ~33.54s to ~5.14s omm.

One backward incompatibility is `Integer::primeFactorization` will now return `[]` (https://en.wikipedia.org/wiki/Empty_product) as the prime factorization of 1 instead of throwing an exception.

### DIFF
--- a/src/NumberTheory/Integer.php
+++ b/src/NumberTheory/Integer.php
@@ -110,12 +110,12 @@ class Integer
      *
      * @return array of prime factors
      *
-     * @throws Exception\OutOfBoundsException if n is < 2.
+     * @throws Exception\OutOfBoundsException if n is < 1.
      */
     public static function primeFactorization(int $n): array
     {
-        if ($n < 2) {
-            throw new Exception\OutOfBoundsException("n must be ≥ 2. ($n provided)");
+        if ($n < 1) {
+            throw new Exception\OutOfBoundsException("n must be ≥ 1. ($n provided)");
         }
 
         $remainder = $n;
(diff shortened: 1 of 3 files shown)


markrogoyski:
Thank you for the pull request adding additional number theory functionality. Looks like good stuff, and thank you for providing unit tests.

I noticed that you added some additional commits since submitting the pull request. Let me know when you think you are done and it is ready for review.

Thanks!

Theriault:
@markrogoyski You're welcome! I think this branch is ready for review. I have additional functionality I am working on but will commit it to a separate branch when I get the time to finish it. Thanks!

Beakerboy:
I’d suggest that you make a list of the features you plan on working on in the issues queue. That way even if you don’t get around to doing it yourself we have an idea of what features people are interested in seeing.

src/NumberTheory/Integer.php
```diff
@@ -23,16 +24,240 @@ public static function isPerfectNumber(int $n): bool
...
+     *    Jₖ(n)
+     *
+     *    (when k=1 - Euler's totient)
+     *    ϕ(n)  φ(n)  phi(n)
+     *
+     * @see    https://en.wikipedia.org/wiki/Euler%27s_totient_function
+     * @see    https://en.wikipedia.org/wiki/Jordan%27s_totient_function
+     *
+     * @param  int $n
+     *
```
markrogoyski: Add an `@param int $k` here with a brief description of what k is.

src/NumberTheory/Integer.php
```diff
@@ -23,16 +24,240 @@ public static function isPerfectNumber(int $n): bool
...
+        return $n - self::totient($n);
+    }
+
+    /**
+     * Reduced totient function (Carmichael function, least universal exponent function)
+     * Return the exponent of the multiplicative group of integers modulo n.
+     *
+     * Notation:
+     *    λ(n)
+     *
```
markrogoyski: How about adding an `@see https://en.wikipedia.org/wiki/Carmichael_function` here for reference.

src/NumberTheory/Integer.php
```diff
@@ -23,16 +24,240 @@ public static function isPerfectNumber(int $n): bool
...
+     * A squarefree integer is an integer which is divisble by no square number other than 1.
+     * It is equal to its radical (squarefree kernel).
+     *
+     * @see    https://en.wikipedia.org/wiki/Square-free_integer
+     * @see    https://oeis.org/A005117
+     *
+     * @param  int $n
+     * @return bool true if n is a squarefree integer; false otherwise
+     */
+    public static function isSquarefreeInteger(int $n): bool
```
markrogoyski: Since the class name is already `Integer`, I think you are OK with just naming this method `isSquarefree`, as it will read `Integer::isSquarefree`.

markrogoyski:
Hi @Theriault,

It looks good. Just a few minor suggestions. Take a look and see what you think.

Thank you for following the unit test style and patterns. The `is` and `isNot` test cases are great.

For a follow-up pull request, consider writing unit tests for axioms and properties. There already is a `NumberTheory\IntegerAxiomsTest` test file. There is not much in there at the moment as the Integer NumberTheory class was quite small before this pull request. For examples of axiom unit tests, take a look at some other axiom unit tests:
* `LinearAlgebra\MatrixAxiomsTest`
* `Number\ComplexNumberAxiomsTest`
* `Probability\CombinatoricsAxiomsTest`
* `Arithmetic\ArithmeticAxiomsTest`

Thanks again for your pull request and new features.

src/NumberTheory/Integer.php
Theriault: Ah good idea!

src/NumberTheory/Integer.php
Theriault: Good catch! I cleaned that up and the missing `@param` from the unit tests as well. I also added a constraint  `k ≥ 1` with unit tests

markrogoyski:
Looks good. Thanks again for the pull request.

PR #361, opened 2019-12-09
Algebra: Refactor Algebra::factors

Theriault:
Utilize `NumberTheory::primeFactorization` to speed up factorization: ~15s vs ~34s to factorize 0..1e6.

Also, fixes squares from returning duplicate factors: `Algebra::factors(4)` used to return `[1,2,2,4]` instead of `[1,2,4]`

### DIFF
--- a/src/Algebra.php
+++ b/src/Algebra.php
@@ -4,6 +4,7 @@ namespace MathPHP;
 
 use MathPHP\Number\Complex;
 use MathPHP\Functions\Map\Single;
+use MathPHP\NumberTheory\Integer;
 
 class Algebra
 {
@@ -110,10 +111,13 @@ class Algebra
      * The decomposition of a composite number into a product of smaller integers.
      * https://en.wikipedia.org/wiki/Integer_factorization
      *
-     * Method:
-     *  Iterate from 1 to √x
-     *  If x mod i = 0, it is a factor
-     *  Furthermore, x/i is a factor
+     * Algorithm:
+     * - special case: if x is 0, return [\INF]
+     * - let x be |x|
+     * - push on 1 as a factor
+     * - prime factorize x
+     * - build sets of prime powers from primes
+     * - push on the product of each set
      *
      * @param  int $x
      * @return array of factors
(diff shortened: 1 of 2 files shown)


markrogoyski:
Hi @Theriault,

Thanks for the pull request and continuing to make MathPHP better!

I ran some tests and verified that these improvements do significantly speed up the factorization process of very large numbers. Does this algorithm have a name? Is it in the list of factoring algorithms on this page: https://en.wikipedia.org/wiki/Integer_factorization ?

Thanks.

Theriault:
The algo doesn't have a name AFAIK. It banks on the prime factorization. As the prime factorization algo improves, this algo will benefit. 

- Prime factorization
- Build a set of prime powers for each distinct prime number. Prime powers here are the most lenient definition including p⁰ and p¹ - see http://oeis.org/wiki/Prime_powers.
- Take the n-ary Cartesian product of those sets
- The product of each set is a factor

As an example, the prime factorization of 360 is 2×2×2×3×3×5 (2³×3²×5¹)

For each distinct prime number, generate a set containing its prime powers:

```
{2⁰, 2¹, 2², 2³}
{3⁰, 3¹, 3²}
{5⁰, 5¹}
```

Take the n‐ary Cartesian product of those sets of prime powers to generate all factors:

```
{2⁰, 2¹, 2², 2³} × {3⁰, 3¹, 3²} × {5⁰, 5¹}
```

Each set generated from the n-ary Cartesian product is a unique factor when you take its product:

```
{2⁰, 3⁰, 5⁰} = {1} = 1
{2¹, 3⁰, 5⁰} = {2} = 2
...
{2², 3², 5¹} = {4, 9, 5} = 180
{2³, 3², 5¹} = {8, 9, 5} = 360
```

That's it. I thought of utilizing `SetTheory\Set` here to make the code a bit cleaner, but it only supports binary Cartesian product at the moment - not n-ary. Also, 2⁹×3⁹×5⁹ would generate 1,000 sets which may be a bit resource intensive compared to iterating 3 arrays of 10 elements each. And a primorial such as 47# would generate thousands of sets. Maybe a variation of Cartesian product for `SetTheory\Set` could be a generator instead. Thanks!

markrogoyski:
Thanks for the explanation.

PR #362, opened 2019-12-13
SetTheory: n-ary Cartesian Product

Theriault:
Add ability to do n-ary Cartesian products

PR #365, opened 2019-12-22
Set-Theory: n-ary Cartesian product

Theriault:
https://github.com/markrogoyski/math-php/pull/362 Merged into develop instead of master.


markrogoyski:
Thanks for the pull request. Sorry for the delay in merging. I was on vacation for the holidays. Thanks again!

PR #373, opened 2020-03-11
Multivariate Hypergeometric

Beakerboy:


Beakerboy:
I have to push some more coverage, but look it over and offer suggestions.

src/Probability/Distribution/Multivariate/Hypergeometric.php
```diff
@@ -0,0 +1,67 @@
...
+
+namespace MathPHP\Probability\Distribution\Multivariate;
+
+use MathPHP\Probability\Combinatorics;
+use MathPHP\Exception;
+
+/**
+ * Multivariate Hypergeometric distribution (multivariate)
+ *
+ * https://en.wikipedia.org/wiki/Multinomial_distribution
```
markrogoyski: I think this is the distribution you are implementing:
https://en.wikipedia.org/wiki/Hypergeometric_distribution#Multivariate_hypergeometric_distribution

tests/Probability/Distribution/Multivariate/HypergeometricTest.php
```diff
@@ -0,0 +1,14 @@
...
+
+use MathPHP\Probability\Distribution\Multivariate\Hypergeometric;
+
+class HypergeometricTest extends \PHPUnit\Framework\TestCase
+{
+    public function testHypergeometric()
+    {
+        $dist = new Hypergeometric([15, 10, 15]);
+        $this->assertEquals(496125 / 3838380, $dist->pmf([2, 2, 2]));
+    }
```
markrogoyski: Let's add the test case from Wikipedia:
```php
        $dist = new Hypergeometric([5, 10, 15]);
        $this->assertEquals(0.079575596816976, $dist->pmf([2, 2, 2]));
```

src/Probability/Distribution/Multivariate/Hypergeometric.php
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+        $picks = array_values($picks);
+
+        $n       = array_sum($picks);
+        $total   = array_sum($this->quantities);
+        $product = 1;
+
+        foreach ($picks as $i => $pick) {
+            $product *= Combinatorics::combinations($this->quantities[$i], $pick);
```
markrogoyski: An alternate way to write this that doesn't involve initializing or side effects.
```php
        $product = array_product(array_map(
            function (int $quantity, int $pick) {
                return Combinatorics::combinations($quantity, $pick);
            },
            $this->quantities,
            $picks
        ));
```
Debatable which is better. Functional patterns in PHP are fairly verbose. But in general, avoiding mutable state is preferable if you can do it without increasing complexity.

markrogoyski:
Thanks for researching the feature request and implementing this quickly. Looks good. Add more unit tests. Maybe implement mean and variance if it is not too much trouble.

src/Probability/Distribution/Multivariate/Hypergeometric.php
Beakerboy: Yes, I was planning on taking care of that. Done.

tests/Probability/Distribution/Multivariate/HypergeometricTest.php
Beakerboy: Done

src/Probability/Distribution/Multivariate/Hypergeometric.php
Beakerboy: We’re the same age, but I’m more old-school...done.

src/Probability/Distribution/Multivariate/Hypergeometric.php
markrogoyski: Maybe you forgot to check it in? Still shows multinomial.

src/Probability/Distribution/Multivariate/Hypergeometric.php
markrogoyski: Well played.

(However, I think Lisp predates C by about 14 years.)

src/Probability/Distribution/Multivariate/Hypergeometric.php
Beakerboy: Odd...now it’s done.

src/Probability/Distribution/Multivariate/Hypergeometric.php
Beakerboy: You’re right, the classics are always in style.

markrogoyski:
LGTM. Any other updates planned or OK to merge?

Beakerboy:
Go ahead and merge.

markrogoyski:
If the quantities are large, the pmf will produce a NAN. This is due to the combinations method relying on factorial which comes up with INF and then dividing INF by INF. R had no problems with large quantities, so I imagine they are computing combinations differently. Something to look into.

Beakerboy:
I see an easy way the combinations function can be improved to increase the upper limit. The first thing it does is to calculate the factorial of n, But later it divides by the factorial of k. Instead, we should calculate the combined product of all the numbers from k plus one through n, inclusive.

PR #376, opened 2020-03-13
More efficient Combination algorithm

Beakerboy:
Permutation can be improved similarly. This will help hypergeometric. instead of being limited by n! overflowing , it's limited by n! / max(n-k,k)! overflowing

### DIFF
--- a/src/Probability/Combinatorics.php
+++ b/src/Probability/Combinatorics.php
@@ -239,16 +239,13 @@ class Combinatorics
             throw new Exception\OutOfBoundsException('k cannot be larger than n.');
         }
 
-        $n！ = self::factorial($n);
-
         // nPn: permutations of n things, taken n at a time
         if (is_null($k)) {
-            return $n！;
+            return static::factorial($n);
         }
 
         // nPk: Permutations of n things taking only k of them
-        $⟮n − k⟯！ = self::factorial($n - $k);
-        return $n！ / $⟮n − k⟯！;
+        return static::fallingFactorial($n, $k);
     }
 
     /**
--- a/src/Probability/Combinatorics.php
+++ b/src/Probability/Combinatorics.php
@@ -287,17 +287,23 @@ class Combinatorics
 
         // nC'k with repetition
         if ($repetition) {
-            $⟮n ＋ k − 1⟯！ = self::factorial($n + $k - 1);
-            $⟮n − 1⟯！k！   = self::factorial($n - 1) * self::factorial($k);
-
-            return $⟮n ＋ k − 1⟯！ / $⟮n − 1⟯！k！;
+            $max = max($n - 1, $k);
+            $min = min($n - 1, $k);
+            $numerator = 1;
+            for ($i = $max + 1; $i <= $n + $k - 1; $i++) {
+                $numerator *= $i;
+            }
+            return $numerator / self::factorial($min);
         }
 
         // nCk without repetition
-        $n！        = self::factorial($n);
-        $⟮n − k⟯！k！ = self::factorial($n - $k) * self::factorial($k);
-
-        return $n！ / $⟮n − k⟯！k！;
+        $max = max($n - $k, $k);
+        $min = min($n - $k, $k);
+        $numerator = 1;
+        for ($i = $max + 1; $i <= $n; $i++) {
+            $numerator *= $i;
+        }
+        return $numerator / self::factorial($min);
     }
 
     /**


markrogoyski:
This is the Hypergeometric PMF test case that I was testing:
```php
            [
                [143, 112, 1889],
                [50, 50, 800],
                0.001039409,
            ],
```
For reference, from R:
```R
> dmvhyper(c(50,50,800), c(143,112,1889), 900)
[1] 0.001039409
```
Even with this iterative method, the combinations still gets out of control:
```
        for ($i = $max + 1; $i <= $n; $i++) {
            $numerator *= $i;
            print("$i numerator:");var_dump($numerator);
        }
```
```
1342 numerator:float(8.7685504128409E+304)
1343 numerator:float(1.1776163204445E+308)
1344 numerator:float(INF)
```

Beakerboy:
I guess the simplest thing would be to just translate the python code in R Into PHP

markrogoyski:
I think the answer might be to use ArbitraryInteger.

Beakerboy:
The RationalNumber class would have to be extended to use ArbitraryInterger as the base number type, and we would have to efficiently convert it `toFloat()`

markrogoyski:
Without using a better algorithm, just using ArbitraryInteger in the same fashion if n! went overboard seemed to work. Although it is significantly slower.

```php
        // nCk without repetition
        $n！        = self::factorial($n);
        $⟮n − k⟯！k！ = self::factorial($n - $k) * self::factorial($k);

        if ($n！ === INF) {
            $bigInt = new ArbitraryInteger($n);
            $n！ = $bigInt->fact();

            $bigInt = new ArbitraryInteger($n - $k);
            $⟮n − k⟯！ = $bigInt->fact();

            $bigInt = new ArbitraryInteger($k);
            $k！ = $bigInt->fact();

            $⟮n − k⟯！k！ = $⟮n − k⟯！->multiply($k！);

            $combinations = $n！->intdiv($⟮n − k⟯！k！);
            return $combinations->toFloat();
        }

        return $n！ / $⟮n − k⟯！k！;
```

Then these large unit tests are all passing.
```php
    /**
     * @test         combinations
     * @dataProvider dataProviderForCombinations
     * @param        int   $n
     * @param        int   $r
     * @param        float $expected
     * @throws       \Exception
     */
    public function testCombinations(int $n, int $r, float $expected)
    {
        // When
        $combinations = Combinatorics::combinations($n, $r);

        // Then
        $this->assertEquals($expected, $combinations, '', 0.00001);
    }

    /**
     * @return array [n, r, combinations]
     */
    public function dataProviderForCombinations(): array
    {
        return [
            [ 170, 6, 30663442810],
            [ 300, 5, 19582837560 ],
            [ 300, 6, 962822846700],
            [ 300, 7, 40438559561400],
            [ 300, 10, 1398320233241701770],
            [ 400, 10, 25798075602615553160],
            [ 400, 20, 2788360983670896737872851072994080],
            [ 500, 30, 1445259588120573502872111127079547366704414552400],
        ];
    }
```

There is probably a way to do this smarter but maybe just as a backup instead of failing, it can be slow but at least correct.

Any thoughts on what the ideal way to do this with Arbitrary Numbers is, knowing that it still needs to return a PHP primitive number type?

Beakerboy:
As I’m thinking about it, this algorithm change does not improve the upper bound on the output as much as the upper bound on the input. This change is worth merging just for that reason.

Here’s how to calculate the maximum possible combination without using the ArbitraryInteger class. Since we know the final answer will be an int, and all ints can be represented as a product of prime factors, we can have an array of factors for each number. Multiplying two numbers will sum the corresponding elements of the array. Dividing subtracts one from the other. In the end, multiply the factors and get a result. If the final answer is able to be represented as a native integer, this algorithm will not overflow at any step.

Beakerboy:
Oops

Beakerboy:
I think I have another good way to avoid overflowing. Instead of creating the numerator, then dividing by the factorial of the denominator, we can test if the new $i within my for loop is divisible by the smallest ‘available’ value from the denominator. If so, divide before adding it into the factorial stack. If not, try the same with the current factorial product before adding in the new value. In theory, all the factors from the denominator should be able to be cancelled out by values in the numerator while building the completed final value.

Beakerboy:
I was expecting this to be the even better inprovement on combinations. All the combination tests pass with it, but there are then failures with some of the discrete distributions:

```php
 // nCk without repetition
// To remove ALL possibility of overflow, each $smallest_min should
// be broken down into prime factors, and consumed before
// multiplying $combination with $factor.
        $max = max($n - $k, $k);
        $min = min($n - $k, $k);
        $combinations = 1;
        $smallest_min = 1;
        for ($i = $max + 1; $i <= $n; $i++) {
            $factor = $i;
            while ($factor % $smallest_min === 0 && $smallest_min <= $min) {
                $factor /= $smallest_min;
                ++$smallest_min;
            }
            $combinations *= $factor;
            while ($combinations % $smallest_min === 0 && $smallest_min <= $min) {
                $combinations /= $smallest_min;
                ++$smallest_min;
            }
        }
        return $combinations;
```

Beakerboy:
Ah, it’s probably less numerically stable once operating on floats due to the extra division operations, and that modulus is inaccurate on floats.

markrogoyski:
There is a floating point version of the remainder function in PHP. Maybe that will help in this case.
https://www.php.net/manual/en/function.fmod.php

markrogoyski:
Is this PR a duplicate now of the one just merged? (https://github.com/markrogoyski/math-php/pull/399)

This also had some changes to permutations. Should this be closed or changed to focus on permutations only?

Beakerboy:
I’ll close this and improve permutations in the same way.

Issue #377, opened 2020-05-07
manhatten & euclidean distance / angle


nessor:
Maybe I am blind, but aren't the functions for determining the manhatten and euclidean distance and the angle between two vectors missing in vector calculus?Then I would implement them and open a PR

markrogoyski:
Hi @nessor,

Thank you for your interest in MathPHP and wanting to improve it.

There is not yet a formal section for Geometric functions dealing with points and shapes and angles, etc. It's on the list of things to do eventually. I have not given a lot of thought to how to model these things yet, so if you have a proposal, please don't hesitate to let us know what you are thinking.

Thanks,
Mark

nessor:
Hi Mark,

I'm thinking of sth. like this:
```
$A = new Vector([1, 2]);
$B = new Vector([2, 4]);

$A->manhattanDistance($B) // 3
$A->euclideanDistance($B) // 2,23607

$A->angle($B) // 0
```

with an implementation of this:
![image](https://user-images.githubusercontent.com/3463261/81373275-9f7ab980-90fc-11ea-97de-11198cf06bf8.png)

PR #378, opened 2020-05-11
Add: Angle and distances between two vectors

nessor:
Hey @markrogoyski,

here are my suggested changes regarding #377.
Since this is my first "real" contribution to an Open Source project, please be nice to me :-D

-Florian

### DIFF
--- a/src/LinearAlgebra/Vector.php
+++ b/src/LinearAlgebra/Vector.php
@@ -222,44 +222,33 @@ class Vector implements \Countable, \Iterator, \ArrayAccess, \JsonSerializable
     /**
      * Calculates the angle between two vectors with
      *
-     *            Σ (x_i * y_i)
-     * ----------------------------------
-     * sqrt( Σ(x_i^2) ) * sqrt( Σ(y_i)^2)
+     *           A⋅B
+     * cos α = -------
+     *         |A|⋅|B|
      *
      * @param Vector $B
      * @return float The angle between the vectors in radians
      * @throws Exception\VectorException
      */
-    public function radAngle(Vector $B)
+    public function cosineSimilarity(Vector $B)
     {
         if (count(array_unique($this->A)) === 1 && end($this->A) === 0)
             throw new Exception\VectorException('The this vector is the null vector');
         else if(count(array_unique($B->A)) === 1 && end($B->A) === 0)
             throw new Exception\VectorException('The parameter vector is the null vector');
 
-        $dotProduct = self::dotProduct($B);
+        $A⋅B     = $this->dotProduct($B);
+        $│A│⋅│B│ = $this->l2Norm() * $B->l2Norm();
 
-        $sqrtSumA = 0;
-        foreach ($this->A as $element){
-            $sqrtSumA += pow($element, 2);
-        }
-        $sqrtSumA = sqrt($sqrtSumA);
-
-        $sqrtSumB = 0;
-        foreach ($B as $element){
-            $sqrtSumB += pow($element, 2);
-        }
-        $sqrtSumB = sqrt($sqrtSumB);
-
-        return acos($dotProduct / ($sqrtSumA*$sqrtSumB));
+        return acos($A⋅B / $│A│⋅│B│);
     }
 
     /**
      * Calculates the angle between two vectors with
      *
-     *            Σ (x_i * y_i)
-     * ----------------------------------
-     * sqrt( Σ(x_i^2) ) * sqrt( Σ(y_i)^2)
+     *           A⋅B
+     * cos α = -------
+     *         |A|⋅|B|
      *
      * @param Vector $B
      * @return float The angle between the vectors in degree
@@ -267,7 +256,7 @@ class Vector implements \Countable, \Iterator, \ArrayAccess, \JsonSerializable
      */
     public function degAngle(Vector $B)
     {
-        return rad2deg(self::radAngle($B));
+        return rad2deg(self::cosineSimilarity($B));
     }
 
     /**
@@ -275,25 +264,12 @@ class Vector implements \Countable, \Iterator, \ArrayAccess, \JsonSerializable
      * https://en.wikipedia.org/wiki/Taxicab_geometry
      *
      * @param Vector $B
-     * @return float|int The manhatten
-     * @throws Exception\VectorException
-     */
-    public function manhattanDistance(Vector $B)
-    {
-        return self::generalDistance($B, 1);
-    }
-
-    /**
-     * Calculates the taxicap geometry between the vectors
-     * https://en.wikipedia.org/wiki/Taxicab_geometry
-     *
-     * @param $B
      * @return float|int
      * @throws Exception\VectorException
      */
-    public function taxicabGeometry($B)
+    public function l1Distance(Vector $B)
     {
-        return self::generalDistance($B, 1);
+        return self::minkowskiDistance($B, 1);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


markrogoyski:
Hi @nessor,

Thank you for the pull request for adding new functionality.

Give me a couple days to become familiar with this and review your code.

One thing I will suggest now is to add a note if possible in your unit tests regarding how you generated the expected test data. For example, if you used an R function or Python function or the examples are from Wikipedia or whatever. It is important to know that you didn't use the code you wrote to generate the test data and that the results can be independently confirmed as correct.

Thanks again for the pull request, and apologies if it takes me more than a couple days to review it.

Mark

nessor:
Hey @markrogoyski,

sure. It's not a problem :) 
I actually "manually" calculated the expected results for the input data using the formulas mentioned in #377 where necessary. Should I still add this as a comment?

Florian

markrogoyski:
Since the Taxicab/Manhattan norm is called `l1Norm` and the Euclidean norm is called `l2Norm` in the Vector class, would it make sense for consistency to call these distance functions `l1Distance` and `l2Distance` instead?

nessor:
You are right, since there is also the l1Norm/l2Norm. Should we perhaps refer to the "names" in the Readme? At least I often search with Ctrl+F (especially with larger readmes) for names/functions I know.
In the code itself I am of your opinion.

nessor:
@markrogoyski 
Should I adjust the code yet, or do you want to have a complete look first?

markrogoyski:
If you have changes go ahead and make them. I still have to take a closer look at some things and look at the unit test cases. Apologies for the delay. The README and the method docblocks are good places to put the alternate names.

src/LinearAlgebra/Vector.php
```diff
@@ -217,6 +219,120 @@ public function perpDotProduct(Vector $B)
         return $A⊥->dotProduct($B);
     }
 
+    /**
+     * Calculates the angle between two vectors with
+     *
+     *            Σ (x_i * y_i)
+     * ----------------------------------
+     * sqrt( Σ(x_i^2) ) * sqrt( Σ(y_i)^2)
```
markrogoyski, line 227: I think if use use the general formula for this that shows the operations in vector terms you would come up with a simpler solution, rather than showing the implementation details in the formula.

For example:
```php
     *           A⋅B
     * cos α = -------
     *         |A|⋅|B|
```

src/LinearAlgebra/Vector.php
```diff
@@ -217,6 +219,120 @@ public function perpDotProduct(Vector $B)
...
+        foreach ($this->A as $element){
+            $sqrtSumA += pow($element, 2);
+        }
+        $sqrtSumA = sqrt($sqrtSumA);
+
+        $sqrtSumB = 0;
+        foreach ($B as $element){
+            $sqrtSumB += pow($element, 2);
+        }
+        $sqrtSumB = sqrt($sqrtSumB);
```
markrogoyski, line 252: I think if you follow the formula, you can arrive at a much simpler solution, which does not leak any implementation details.

```php
        $A⋅B     = $this->dotProduct($B);
        $│A│⋅│B│ = $this->l2Norm() * $B->l2Norm();

        return acos($A⋅B / $│A│⋅│B│);
```

src/LinearAlgebra/Vector.php
```diff
@@ -217,6 +219,120 @@ public function perpDotProduct(Vector $B)
...
+     * Calculates the general distance between vectors with
+     *
+     * ( Σ|x_i - y_i|^p )^1/p
+     *
+     * @param Vector $B
+     * @param int $p
+     * @return float|int
+     * @throws Exception\VectorException
+     */
+    public function generalDistance(Vector $B, int $p)
```
markrogoyski, line 322: Would `minkowskiDistance` be a better name for this method?
https://en.wikipedia.org/wiki/Minkowski_distance

src/LinearAlgebra/Vector.php
```diff
@@ -217,6 +219,120 @@ public function perpDotProduct(Vector $B)
...
+     *
+     *            Σ (x_i * y_i)
+     * ----------------------------------
+     * sqrt( Σ(x_i^2) ) * sqrt( Σ(y_i)^2)
+     *
+     * @param Vector $B
+     * @return float The angle between the vectors in radians
+     * @throws Exception\VectorException
+     */
+    public function radAngle(Vector $B)
```
markrogoyski, line 233: Would `cosineSimilarity` be a better name for this method?
https://en.wikipedia.org/wiki/Cosine_similarity

src/LinearAlgebra/Vector.php
nessor: You're absolutely right. I'll adjust that.

src/LinearAlgebra/Vector.php
nessor: Again, you're right :)

nessor: Any ideas for the degAngle method? Or actually remove them?

src/LinearAlgebra/Vector.php
nessor: I think its good! I adjusted it :)

src/LinearAlgebra/Vector.php
nessor: Implemented.

src/LinearAlgebra/Vector.php
markrogoyski: How about only having the `cosignSimilarity` method, but having an optional second parameter to return degrees.

One option is something like this:
```php
public function cosignSimilarity(Vector $B, inDegrees = false): float { ... }
```

markrogoyski:
Hi @nessor,
Any update on this?

Thanks,
Mark

nessor:
> Hi @nessor,
> Any update on this?
> 
> Thanks,
> Mark

Hey @markrogoyski,

no. I'm afraid I haven't gotten around to it yet. The last few days have been a little crowded. I'll try to check it next weekend :)

markrogoyski:
Hi @nessor,

No worries. No rush. Getting things right is more important than getting them done fast. Just wanted to make sure this was still on your radar.

Thanks for the update.
Mark

nessor:
@markrogoyski 
Sorry for the delay. I don't know what's going on right now. Somehow everybody wants something from me... But I pushed the few changes :)

markrogoyski:
Hi @nessor,

Thanks for making the changes. Let me know if you are planning more changes or if it is ready to merge.

Thanks for the update.
Mark

nessor:
Hi @markrogoyski ,

I still have some ideas, but unfortunately no time at the moment :(
Since I will use the package in my project myself, I will implement the missing functions and create a new PR :)

markrogoyski:
@nessor,

Is this OK to merge? I'll make any adjustments if necessary if you are OK with me merging it as is.

Thanks.
Mark

nessor:
@markrogoyski 

> @nessor,
> 
> Is this OK to merge? I'll make any adjustments if necessary if you are OK with me merging it as is.
> 
> Thanks.
> Mark

Sure :)

markrogoyski:
Hi @nessor,

Thanks again for the pull request. Your changes have been merged and incorporated into the latest release, v1.3.0 (https://github.com/markrogoyski/math-php/releases/tag/v1.3.0).

I made some changes, moving the core distance functionality into the existing `Statistics\Distance` namespace and having the `Vector` class pass through to those.

Thanks again. And apologies it took some time to get to this pull request.

nessor:
Hey @markrogoyski,

thanks for the first guided PR :) Feels good to have done something for the community! Will not be the last PR :o)

PR #385, opened 2020-09-25
PLS

Beakerboy:
I've created a PLS class that calculates all the standard matrices used for PLS regression. R has a few different PLS implementations in different libraries. the mvr() function appears to be the most mature, but does a lot of optimization behind the scenes. the chemometrics/pls2_nipals() function looks like a textbook example of the algorithm and provides results that can be used to test against.


Beakerboy:
There are a few more tests that need to be added to the PLS1 test case, and I need a test for the `predict()` method.
Also, do you think more descritive get function names would be better? R's `pls/mvr()` uses:

- coefficients for B
- Yloading for C
- projection for P
- score for T
- Yscore for U
- loading.weight for W

markrogoyski:
Hi @Beakerboy,

Thanks for the pull request of this recently requested feature.

I need to learn a little bit more about this, but my initial reaction is that names like `coefficients` and `projection` are better than `B` and `P`.

markrogoyski:
One options is to use short names for the internal implementation and have the public interface getters use the descriptive names.

Beakerboy:
> have the public interface getters use the descriptive names.

Exactly what I was thinking

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+                $t = $E->multiply($w);
+                $c = $F->transpose()->multiply($t)->scalarDivide($t->frobeniusNorm() ** 2);
+                $new_u = $F->multiply($c);
+                $diff = $new_u->subtract($u)->frobeniusNorm();
+            } while ($diff > $tol);
+            $u = $new_u;
+
+            // Least squares regression on a slope-only model: 𝜷ᵢ = Σ(xᵢyᵢ) / Σ(xᵢ²)
+            $p = $E->transpose()->multiply($t)->scalarDivide($t->frobeniusNorm() ** 2);
+            $q = $F->transpose()->multiply($u)->scalarDivide($u->frobeniusNorm() ** 2);
```
markrogoyski, line 103: `$q` is not used anywhere.

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+            $E = $E->subtract($t->multiply($p->transpose()));
+            $F = $F->subtract($t->multiply($c->transpose())->scalarMultiply($d));
+
+            // Add each of these columns to the overall matrices
+            $this->C = is_null($this->C) ? $c : $this->C->augment($c);
+            $this->P = is_null($this->P) ? $p : $this->P->augment($p);
+            $this->T = is_null($this->T) ? $t : $this->T->augment($t);
+            $this->U = is_null($this->U) ? $u : $this->U->augment($u);
+            $this->W = is_null($this->W) ? $w : $this->W->augment($w);
+            $this->D[] = $d;
```
markrogoyski, line 116: The `D` collection does not seem to be used anywhere.

src/Statistics/Multivariate/PLS.php
Beakerboy: The R package that the algorithm comes from has this as an output, but it seems not to be common to subject this matrix to analysis. I can't find a good name for it.

src/Statistics/Multivariate/PLS.php
Beakerboy: The R package that the algorithm comes from has this as an output, but it seems not to be common to subject this matrix to analysis. I can't find a good name for it. It's a diagonal matrix, hence 'D'. I have it here in case you or someone else says it's important.

tests/Statistics/Multivariate/PLS1ScaleFalseTest.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+namespace MathPHP\Tests\Statistics\Multivariate;
+
+use MathPHP\LinearAlgebra\Matrix;
+use MathPHP\LinearAlgebra\MatrixFactory;
+use MathPHP\SampleData;
+use MathPHP\Statistics\Multivariate\PLS;
+
+class PLS1ScaleFalseTest extends \PHPUnit\Framework\TestCase
```
markrogoyski, line 10: Should this be called PLS2?

Beakerboy: PLS1 is a simpler algorithm that works only when y is a single column. Multiple Y values requires PLS2. If we decided to add separate PLS1 and PLS2 algorithms, allowing the PLS class to choose the best algorithm, then that file could be used to test the PLS1 implementation.

tests/Statistics/Multivariate/PLS1ScaleFalseTest.php
```diff
@@ -0,0 +1,86 @@
...
+            [-0.0002882169],
+        ];
+
+        // When
+        $B = self::$pls->getCoefficients()->getMatrix();
+
+        // Then
+        $this->assertEquals($expected, $B, '', .00001);
+    }
+}
```
markrogoyski, line 86: Can we add unit tests for all the other C, P, T, U, W, etc. variables as well?

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+use MathPHP\Exception;
+use MathPHP\LinearAlgebra\Matrix;
+use MathPHP\LinearAlgebra\MatrixFactory;
+use MathPHP\LinearAlgebra\Vector;
+use MathPHP\Statistics\Descriptive;
+
+/**
+ * Partial Least Squares Regression
+ *
+ * Using the NIPALS PLS2 algorithm
```
markrogoyski, line 14: Can you link me to the source you used to base this implementation on so I can read up on it? Thanks.

src/Statistics/Multivariate/PLS.php
markrogoyski: How about just adding it as `D` then? At a minimum, it is some extra validation for unit tests.

src/Statistics/Multivariate/PLS.php
Beakerboy: https://rdrr.io/cran/chemometrics/src/R/pls2_nipals.R

tests/Statistics/Multivariate/PLS1ScaleFalseTest.php
Beakerboy: Can do

tests/Statistics/Multivariate/PLS1ScaleFalseTest.php
```diff
@@ -0,0 +1,86 @@
...
+     *   library(chemometrics)
+     *   X = mtcars[,c(2:7, 10:11)]
+     *   Y = mtcars[,c(1)]
+     *   pls.model = pls2_nipals(X, Y, 1)
+     *
+     * @throws Exception\MathException
+     */
+    public static function setUpBeforeClass()
+    {
+        $mtCars = new SampleData\MtCars();
```
markrogoyski, line 32: This is really nice, using the sample data for the unit tests. Makes it easy to set up in both R and here.

src/Statistics/Multivariate/PLS.php
markrogoyski: Thanks!

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+
+    /** @var Matrix $T X Scores*/
+    private $T = null;
+
+    /** @var Matrix $U Y Scores*/
+    private $U = null;
+
+    /** @var Matrix $W X Weights*/
+    private $W = null;
+
```
markrogoyski, line 49: Missing `private $D` as it is being assigned to later on `$this->D[] = $d;`.

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+    private $Ycenter;
+ 
+    /** @var Vector X Scale */
+    private $Xscale;
+
+    /** @var Vector Y Scale */
+    private $Yscale;
+
+    /** @var Matrix $B Regression Coefficients*/
+    private $B = null;
```
markrogoyski, line 33: These null initializations are not necessary. Null is the natural state of an uninitialized property.
```php
php > class MyClass {
php {   private $thing;
php { }
php > $myClass = new MyClass();
php > var_dump($myClass);
object(MyClass)#1 (1) {
  ["thing":"MyClass":private]=>
  NULL
}
```

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+     *  - getU
+     *  - getW
+     **************************************************************************/
+
+    /**
+     * Get the regression coefficients
+     *
+     * The matrix that best transforms E into F
+     */
+    public function getCoefficients()
```
markrogoyski, line 139: I think you can safely add return type declarations to all of these getters: `: Matrix`

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Predict Values
+     *
+     * Use the regression model to predict new values of Y given values for X.
+     * Y = (X - μₓ) ∗ σₓ⁻¹ ∗ B ∗ σ + μ
+     */
+    public function predict(Matrix $X)
+    {
+        if ($X->getN() !== $this->Xcenter->getN()) {
+            throw new Exception\BadDataException('Data does not have the correct number of columns.');
```
markrogoyski, line 203: Add this unit test to trigger this error:
```php
    /**
     * @test predict error if the input X columns do not match
     */
    public function testPredictDataColumnMisMatch()
    {
        // Given
        $X = MatrixFactory::create([[6, 160, 3.9, 2.62, 16.46]]);

        // Then
        $this->expectException(\MathPHP\Exception\BadDataException::class);

        // When
        $prediction = self::$pls->predict($X);
    }
```

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+     * @param int     $ncomp number of components to use in the model
+     * @param boolean $scale standardize each column?
+     *
+     * @throws Exception\BadDataException if any rows have a different column count
+     */
+    public function __construct(Matrix $X, Matrix $Y, int $ncomp, bool $scale = false)
+    {
+        // Check that X and Y have the same amount of data.
+        if ($X->getM() !== $Y->getM()) {
+            throw new Exception\BadDataException('X and Y must have the same number of rows.');
```
markrogoyski, line 64: Add a test like t his to trigger this exception:
```php
    /**
     * @test Construction error - row mismatch
     */
    public function testConstructionFailureXAndYRowMismatch()
    {
        // Given
        $Y = self::$Y->rowExclude(0);

        // Then
        $this->expectException(\MathPHP\Exception\BadDataException::class);

        // When
        $pls = new PLS(self::$X, $Y, 2, true);
    }
```

src/Statistics/Multivariate/PLS.php
```diff
@@ -0,0 +1,252 @@
...
+        $center_matrix = $center_matrix ?? $ones_column->multiply(MatrixFactory::create([$center->getVector()]));
+
+        // Create a diagonal matrix of the inverse of each column standard deviation.
+        $scale_matrix = MatrixFactory::diagonal($scale->getVector())->inverse();
+
+        // scaled data: ($X - μ) ∗ σ⁻¹
+        return $new_data->subtract($center_matrix)->multiply($scale_matrix);
+    }
+
+    private static function columnStdevs(Matrix $M)
```
markrogoyski, line 244: Can you add a short PHPDoc description of what the intent of this function is?

markrogoyski:
Are there any edge cases to be concerned about that should have tests? Like a Matrix full of zeros for instance?

markrogoyski:
The R chemometrics library has a sample data set that is made specifically for PLS2.
Cereal data set: https://rdrr.io/cran/chemometrics/man/cereal.html
From the documentation: _The data set can be used for PLS2._

I've added this data set to the SampleData in MathPHP in the develop branch.
https://github.com/markrogoyski/math-php/blob/develop/src/SampleData/Cereal.php

I thought I'd try my hand at writing a unit test for this, but it didn't work. It seems to go into an infinite loop.

R code looks like this:
(From the R documentation example for pls2_nipals: https://www.rdocumentation.org/packages/chemometrics/versions/1.4.2/topics/pls2_nipals)
```R
> data(cereal)
> res <- pls2_nipals(cereal$X,cereal$Y,a=5)
```

PHPUnit test starts off like this:
```php
    public function testCerealPls()
    {
        // Given
        $cereal = new SampleData\Cereal();
        $X      = MatrixFactory::create($cereal->getXData());
        $Y      = MatrixFactory::create($cereal->getYData());

        // And
        $pls = new PLS($X, $Y, 5);

        // When
        $B = $pls->getCoefficients()->getMatrix();

        // Then
        // Test something...
    }
```

I left it running for quite some time and it didn't not finish. R finishes almost immediately. However, R does spit out some warnings about iterations without converging:
```R
> res <- pls2_nipals(cereal$X,cereal$Y,a=5)

WARNING! Iteration stop in h= 2  without convergence!
WARNING! Iteration stop in h= 3  without convergence!
WARNING! Iteration stop in h= 4  without convergence!
WARNING! Iteration stop in h= 5  without convergence!
```

It seems to map to this conditional block in the R implementation:
```R
		if (unorm<tol) {
			ende <- TRUE
		}
		else if (nr > it) {	#too many iterations
			ende <- TRUE
			cat("\nWARNING! Iteration stop in h=",h," without convergence!\n\n")
		}
```

Not really understanding the code that much, I just put some `print();ob_flush();` lines here and there and it seems like it got stuck in the do-while loop with this condition never hitting ` } while ($diff > $tol);`.

Can you play around with this dataset in `SampleData\Cereal` and see what iteration checks we might need here? I think it would be great validation of the PLS code to use the same sample data the R chemometrics package provides.

Beakerboy:
It looks like I'll just add the same loop-breaking code to break out if the code does not converge. It's possible that, like some of the eigenvector functions, the resulting vector is alternating signs.

Beakerboy:
I believe all your issues have been addressed.

markrogoyski:
Thanks for revisiting this and making improvements.

I tried writing a unit test using the R Cereal data set again, and now it completes whereas before it did not terminate. However, while the values were similar, they weren't the same. I didn't know if I was doing something wrong or not. Since this data set was intended for PLS calculations, it seemed like a good way to test your PLS implementation since we have the same data set in `SampleData`.

This is what I did in R to get a PLS of the cereal data.
```R
> library(chemometrics)
> data(cereal)
> res <- pls2_nipals(cereal$X, cereal$Y, a=5, scale=FALSE)
```

Do you mind taking a look and see if you can write unit tests based on R's cereal data set PLS calculation?

Thanks.

Beakerboy:
This calculated the T values correctly with the cereal data. I'll add the rest to see which are off.

Beakerboy:
It looks like when you exported the Cereal data, you may not have pulled everything to its max preceision.

```R
library(chemometrics)

data(cereal)
options(digits = 15) 
cereal$X
```
The first value is 0.0026827552 while you have 0.002682755.

markrogoyski:
> It looks like when you exported the Cereal data, you may not have pulled everything to its max preceision.
> 
> ```r
> library(chemometrics)
> 
> data(cereal)
> options(digits = 15) 
> cereal$X
> ```
> 
> The first value is 0.0026827552 while you have 0.002682755.

I probably used the defaults. Is that making a big difference in matching R test results? If so I could remake the SampleData.

Beakerboy:
It's an iterative method, so I imagine the butterfly effect is in effect.

markrogoyski:
Looking at the PLS code, it seems like it is mostly arithmetic, except that it takes a matrix inverse at one point to compute "B". The inverse still uses RREF and could be quite unstable. Perhaps this is a good opportunity to look into rewriting matrix inverse to use QR or LU or something.

Beakerboy:
B is calculated at the end, and even the C and W matrices have precision issues.

Beakerboy:
Given that this data set fails to converge, I'm guessing it is unstable and difficult to get exact results. I don't know exactly how the data is saved in the data file, but it's possible that exporting and importing via text, changes things enough that it does not work.

The library calculates C a little different. there is one method that is commented out, then another. I think I use the commented out version since this is the "textbook" way. The other may be algebraically equivalent, but numerically different.

Beakerboy:
It looks like cereal data works fine when it is scaled.

Beakerboy:
This should be ready to go. It appears the Cereal dataset works best when rescaled.

markrogoyski:
Thanks for revisiting and working on this and looking into using the Cereal data set for unit tests!

PR #387, opened 2020-10-01
math php

Pranay711:
added  linear equation in algebra.


markrogoyski:
Hi @Pranay711,

Thank you for your interest in MathPHP.

Were you planning to add new functionality? I see you added `Algebra::linear` to the README but there is no code to implement this.

Mark

Beakerboy:
This feature request seems trivially simple:

```php
    public static linear($a, $b)
    {
        return -1 * $b / $a;
    }
````

markrogoyski:
@Beakerboy, 
Based on the name it seems like just solving the equation ax +b = 0 like you illustrate.

@Pranay711,
Can you confirm this is the feature you are looking to implement?

Thanks.
Mark

markrogoyski:
This has been implemented and will be released in the next version: v1.7.0. Thank you for suggesting this feature.

PR #390, opened 2020-10-06
Added tests for ArbitraryInteger::multiply() and fixed a bug. 

Beakerboy:
The sign of the input number was not being taken into consideration.


markrogoyski:
OK lol at my comment in the other PR. I guess it was broken. No we have tests for it! yay!

PR #397, opened 2020-10-22
Rational inverse() and  pow()

Beakerboy:


markrogoyski:
Nice. Thanks for adding this. There is a typo in the exception messages. I'll fix it.

PR #399, opened 2020-11-12
Expanding precision range of combination function.

Beakerboy:
Using a falling factorial helps the function return an int when n or n+k is greater than 20.


src/Probability/Combinatorics.php
```diff
@@ -285,19 +285,25 @@ public static function combinations(int $n, int $k, bool $repetition = false): f
...
+            $numerator = $n;
         }
 
-        // nCk without repetition
-        $n！        = self::factorial($n);
-        $⟮n − k⟯！k！ = self::factorial($n - $k) * self::factorial($k);
-
-        return $n！ / $⟮n − k⟯！k！;
+        // The internal falling factorial implementation always returns a float.
+        // Here we maintain int precision as much as possible.
```
markrogoyski, line 299: Does this make a difference? Falling factorial doesn't do any division, and this method does a division at the end anyways.

Beakerboy: The test cases that I added fail without this pull request. The result is necessarily an integer, but without using the falling factorial, the int data type will overflow into a float. The division isn’t the problem, it’s the factorial.

markrogoyski: OK. I see what happens with the result.
`Failed asserting that 3247943159.9999995 matches expected 3247943160.0`

markrogoyski:
Thanks for the pull request to improve combinations.

I think the algorithm you are using is called the multiplicative formula, which does seem to be more efficient and less likely to overflow. Thanks for thinking to make this improvement. Do you mind documenting this in the code and making it clear how the code matches the formula a little better?

For example, for nCk without repetition, I think the formula would be this:
```php
            /*
             * Multiplicative formula
             *
             *        (n)   _ᵏ_  n + 1 - i
             *  nCk = (k) = | |  ---------
             *              ⁱ⁼¹      i
             *
             */
            $∏⟮n ＋ 1 − i⟯ = 1;
            for ($i = 1; $i <= $k; $i++) {
                $∏⟮n ＋ 1 − i⟯ *= ($n + 1 - $i);
            }
            return $∏⟮n ＋ 1 − i⟯ / self::factorial($k);
```
And using your method to only do one division at the end seems to be less likely to cause a precision error than to do it with each iteration of the loop.
I'm not quite sure I understand the min/max trick, so a short explanation of that would be helpful too.

The Wikipedia page I think is showing the formula for combinations without repetitions: https://en.wikipedia.org/wiki/Binomial_coefficient#Multiplicative_formula
What is the formula for with repetitions?

Thanks,
Mark

Beakerboy:
Again, the division is not a problem. Since the answer is always an integer, there will never be remainders when the division takes place, as king as the division takes place as the last step. Improving these functions entirely lies with preventing overflow.

When a factorial is divided by a smaller factorial, all the elements from the smaller will be cancelled out by the corresponding values in the larger element, leaving behind a falling factorial in the numerator. If we want to make the numerator as small as possible, to limit overflow, and you have your choice between two values in the denominator, we want to choose the larger of the two. That’s what the max/min is for.

If you use the multiplicative formula as written with combinations(22, 22) it will return a float ≈ 1 instead of an int === 1. Using the max of the denominator will ensure it remains an int.

markrogoyski:
Thanks for the explanation. I think I get it. Tests all pass so seems good. Do you mind, either as a comment here, or in the code, just write out both of the formulas for with and without repetitions so I can more easily follow the code. Thanks. Sorry for not quite fully understanding this yet.

Beakerboy:
result = A! / B! / C!  = (A! / B!) / C! = (A! / C!) / B! 

if  x > y
                   x
 x! / y!  = Π(i)
                   y+1

If y > z then (x! / y!) < (x! / z!)

If we want all steps to have the smallest integers, A! / B! will be smaller than A!. If A! would normally overflow, since we will be dividing it by B!, we can create the falling factorial instead and skip the division step.

I don’t know how best to explain it. Maybe write out a few examples by hand to see how things cancel, then imagine how they would behave similarly when some of the numbers are 21 or greater. A 64 bit int overflows at 21!.

markrogoyski:
OK. I'm going to merge now and work through it later. Thanks for the explanation. I'll figure it out eventually.

Beakerboy:
If we’re back in the world of 8 bit computers, 7! Is the max integer value. If we wanted to do calculations with n=8 and k=8, we would have 8! / 0! / 8! which is 1. Under the original algorithm, each of these factorials is evaluated separately, to both the 8! values would be floats. Under the new algorithm, we see the max denominator is 8, so that falling factorial immediately evaluates to 1, and we get an int with the value of 1. If we use the Wikipedia algorithm, we will divide by k!; which is 8!...a float, so the int data type is not maintained.

Similarly, if n=9 and k=1, in the original algorithm, with 9! / 1! / 8! We would have overflows. In when using the  new algorithm, the (n-k)! Is largest, and 9! / 8! = 9, so we get the 9 as an int because nothing overflows. This example would work under the Wikipedia algorithm, because k is small. The example above does not because k > 7.

markrogoyski:
Thanks again for the further explanation.

I understand the canceling out. It makes sense. Don't do any unnecessary calculations. Makes sense.

My remaining question is how you determined the numerator and denominator for both with and without repetitions. I can imagine six months from now if someone posts an issue/bug and maybe neither you nor I remember why it is the way it is. That's why I'd like to have some documentation or link to a formula/algorithm or something to correlate with the code. I'll happily write it up if you can point me in the right direction. I put some basic documentation from Wikipedia, but I don't think it is right.

Thanks,
Mark

Beakerboy:
Ahhh. Those were in the original algorithm. Both versions (with and without replacement) have a $k in the denominator. The original algorithm had the other two for each version.

I just noticed that this function does not allow $k to be greater than $n when replacement is true. That’s a bug.

PR #400, opened 2020-11-14
K can be greater than n when repetition is allowed.

Beakerboy:


markrogoyski:
Merge target is master. Change to develop so I can find the actual change in the diff =)

PR #401, opened 2020-11-14
K can be greater than n when repetition allowed

Beakerboy:


markrogoyski:
Thanks for the PR. This has definitely had the wrong behavior for at least four years. Thanks for fixing it!

For reference:
```python
>>> scipy.special.comb(5, 7, repetition=True)
330.0
```

PR #402, opened 2020-11-14
Oops, missing $

Beakerboy:


markrogoyski:
The Travis service is so slow lately. They aren't even waiting for the cut-off date. They seemed to have gotten rid of all the resources. PR builds queue up for hours before running.

Anyways, I will work on moving the CI checks to GitHub Actions and hopefully they will run immediately and catch these kinds of things within minutes rather than hours or days like it has been lately.

PR #404, opened 2020-11-16
Permutation precision error

Beakerboy:
Preventing the factorial from overflowing.

### DIFF
--- a/src/Probability/Combinatorics.php
+++ b/src/Probability/Combinatorics.php
@@ -239,16 +239,17 @@ class Combinatorics
             throw new Exception\OutOfBoundsException('k cannot be larger than n.');
         }
 
-        $n！ = self::factorial($n);
-
         // nPn: permutations of n things, taken n at a time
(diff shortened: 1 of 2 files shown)


markrogoyski:
Thanks for following up and improving permutations as well.

Beakerboy:
I wrote a simple for loop to test all $n less than 100 and all $k < $n to find the first permutation where (float) (int) $perm !== $perm as long as $perm < PHP_INT_MAX. The new test case failed.

markrogoyski:
What were you expecting to happen? And did the previous code fair any better?

Beakerboy:
I’m saying, I expected the old code to have an issue, and this is how I found the test case that would fail.

markrogoyski:
OK I got it. Old code fails in test case `[23,  5, 4037880],` and passes now. Thanks.

For reference:
```
Failed asserting that 4037880.0000000005 matches expected 4037880.0.
```

PR #405, opened 2020-11-16
Use the MatrixFactory in test

Beakerboy:
Instead of creating Matrix directly.


markrogoyski:
Thanks for the pull request. Good change. I'm going to wait for Travis to actually run before I merge. I think I've learned my lesson from merging too early =). BTW, I started looking into Github actions to do all the CI. I'll try to get something set up for pull requests soon.

Beakerboy:
Wait if you want, but for what it’s worth, I ran it on my end before submitting it:
https://travis-ci.org/github/Beakerboy/math-php/jobs/743979703

PR #407, opened 2020-11-16
Introducing the Square matrix trait

Beakerboy:
I’ve migrated a few functions out of the main Matrix class, into a trait. Let me know what you think of this process. Nothing was changed in tests, so I’d consider this not breaking the API.

### DIFF
--- a/src/LinearAlgebra/Matrix.php
+++ b/src/LinearAlgebra/Matrix.php
@@ -1543,173 +1543,31 @@ class Matrix extends MatrixBase
     /**
      * Inverse
      *
-     * For a 1x1 matrix
-     *  A   = [a]
-     *  A⁻¹ = [1/a]
-     *
-     * For a 2x2 matrix:
-     *      [a b]
-     *  A = [c d]
-     *
-     *         1
-     *  A⁻¹ = --- [d -b]
-     *        │A│ [-c a]
-     *
-     * For a 3x3 matrix or larger:
-     * Augment with identity matrix and calculate reduced row echelon form.
-     *
-     * @return Matrix
-     *
      * @throws Exception\MatrixException if not a square matrix
-     * @throws Exception\MatrixException if singular matrix
-     * @throws Exception\IncorrectTypeException
-     * @throws Exception\BadParameterException
-     * @throws Exception\OutOfBoundsException
      */
     public function inverse(): Matrix
     {
-        if ($this->catalog->hasInverse()) {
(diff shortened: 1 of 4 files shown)


markrogoyski:
Some design questions to think about.

Does the SquareMatrixTrait make things easier to reason about or more complex? Are we just trying to remove an if-not-square-throw-exception guard statement or is this part of a larger enabling change that improves the design and simplifies things in multiple other places? What becomes easier/simpler in the long run because of this change? The if-not-square guard clause is pretty straightforward and all the logic is in one place.

What is considered more common: square matrices or non-square matrices? If the answer is square, then would it make sense for the default numeric matrix to do the inverse, and any non-square matrix throw the exception? Would it make sense to have a NumericSquareMatrix class and NonNumericSquareMatrix? Is that complexity better than if-not-square guard?

Beakerboy:
The plan that we had discussed in the past was to move square matrix functions to a trait to allow an ObjectSquareMatrix class with less duplication. It would also make the Matrix class smaller and more focused. Any structural functions, like minorMatrix would work on a square matrix with any data type. Inverse() probably shouldn’t move here since I just noticed a scalar division operation. 

Question, should I keep the stub in the Matrix class that throws an exception or not?

I’d remove the `if (! $this->isSquare())` stuff from the Trait since it will never enter that control structure.

Beakerboy:
This refactor must have decreased lines of code; all changed files are at 100% coverage.

markrogoyski:
Sorry for the delay in reviewing this. Hope you had a great Thanksgiving. Thanks for being a great MathPHP partner!

Just some additional thoughts. I've used traits in other projects and it tends to get confusing to have `$this` in a trait. And more so if there are base methods being overridden by the trait. At least for me, I find it hard to reason about the code and even my IDE gets a bit confused at times.

Another way to consolidate common code could be via composition of other objects, or utility classes with static methods.

For example, maybe the base Matrix class defines something like `inverse` or `minorMatrix` as abstract. Then square matrices could invoke `SquareMatrixOperations::minorMatrix($this)` or `InverseCalculator::calculate($this)`. It would be a bit more code, but there would be less "magic" to figure out what is going on. You still get the benefit of consolidating common code in a single place for reuse, and testing it independently. A non square matrix would specifically have to define a method that throws the exception.

Sometimes a little copy and pasting is better than some programming language magic if it makes the code clearer. What do you think?

markrogoyski:
I refactored the matrix class in the v2.0.0 release.

Issue #356, opened 2019-11-06
IRR function


kman2013:
I'm having an issue with the IRR function
I have a payment stream (same payment amount) for 20 years and then it stops - then I have 0 for 30 years showing and a future value at year 50 - this function returns a negative value (-212,27%) but in Excel I get a value of 5.09%. It correctly gave me a value at year 49.
Is there a max number of values it can do?

markrogoyski:
Hi,
Can you provide your input list of values so I can try to reproduce it?
Thanks.

kman2013:
$values = array(-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,250010);

markrogoyski:
Thank you for the example input. I'm able to reproduce it. I'll update this issue when I have an update.

kman2013:
awesome - thank you so much!!

kman2013:
just to add on to the point - I would like to be able to do the values up to 100 years out - so this one stops at year 49 - so the next 51 years of data wouldn't work - if that is possible... :)

markrogoyski:
Hi @kman2013,

Internal rate of return (IRR) is an interesting function.

To get the result you are looking for that matches what Excel comes up with, provide an initial guess different than the default one. For example:

```php
$values = [-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,-1607,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,250010];
$initialGuess = 0.01;

$irr = Finance::irr($values, $initialGuess); // 0.05090132749
```
This will get the result of 0.05090132749 that you got in Excel.

[...]

PR #412, opened 2020-12-18
Add alternate IRR implementation for NAN handling.

adamrogoyski:
Add alternate stable implementation of Finance IRR function to
compute an answer when the current implementation's initial guess
gives a NAN.

Fix typo: band -> bond.


markrogoyski:
This addresses issue https://github.com/markrogoyski/math-php/issues/356.

PR #417, opened 2021-07-02
Quaternions

Beakerboy:
Needs more tests, but I figure I should get this in the pull request queue.


src/Number/Quaternion.php
```diff
@@ -0,0 +1,345 @@
...
+     * Constructor
+     *
+     * @param number $r Real part
+     * @param number $i Imaginary part
+     * @param number $j Imaginary part
+     * @param number $k Imaginary part
+     */
+    public function __construct($r, $i, $j, $k)
+    {
+        $this->r = $r;
```
markrogoyski, line 55: We probably want to have some sort of check that you don't pass in strings or arrays or something. You want to fail early during construction rather than later on when you do some arithmetic operation which couldn't possibly work. Either add some check (`is_numeric` or `is_int || is_float`) or add a type hint for float in the constructor.

I also noticed the Complex class also has this problem.

BTW, PHP 8 has union types. So in the future we can put `int|float` in the function declaration.

src/Number/Quaternion.php
```diff
@@ -0,0 +1,345 @@
...
+     */
+    public function __toString(): string
+    {
+
+        if ($this->r == 0 & $this->i == 0 & $this->j == 0 & $this->k == 0) {
+            return '0';
+        } else {
+            $i_neg = false;
+            $j_neg = false;
+            $k_neg = false;
```
markrogoyski, line 85: I don't see these three boolean values being used anywhere.

src/Number/Quaternion.php
```diff
@@ -0,0 +1,345 @@
...
+
+            $tr = "$this->r";
+            $ti = "$this->i" . 'i';
+            $tj = "$this->j" . 'j';
+            $tk = "$this->k" . 'k';
```
markrogoyski, line 90: I don't see these four values being used anywhere.

src/Number/Quaternion.php
```diff
@@ -0,0 +1,345 @@
...
+
+            $chk0 = function ($q) {
```
markrogoyski, line 92: Since these functions are not closing over any local scope (no `use` directive), you could just make these private methods of the class rather than variable-named anonymous functions.

src/Number/Quaternion.php
Beakerboy: Thanks. This code was from another contributor, and I’ve just started tweaking it and adding tests. There seems to be an issue with the code coverage reporting through GitHub actions. I was hoping the coverage report would quickly highlight some of these problems.

src/Number/Quaternion.php
```diff
@@ -0,0 +1,345 @@
...
+                } else {
+                    return "$q";
+                }
+            };
+
+            $chk0i = function ($q, $unit) {
+                if ($q == 0) {
+                    return "";
+                } else {
+                    if ($q > 0) {
```
markrogoyski, line 104: The extra else levels of indentation are not needed here. Both early returns allow a single, simpler indentation flow.
```php
if ($q == 0) {
    return "";
}
if ($q > 0) {
    return ' + ' . "$q" . $unit;
}
return ' - ' . (string) \abs($q) . $unit;
```

Beakerboy:
Going to restart

tests/Number/QuaternionAxiomsTest.php
```diff
@@ -0,0 +1,121 @@
...
+ * all work out according to the axioms.
+ *
+ * Axioms tested:
+ *  - Identity
+ *    - z + 0 = z
+ *    - z * 1 = z
+ *  - Inverse
+ *    - (∀a)(∃b) a + b = 0
+ */
+class QuaternionNumberAxiomsTest extends \PHPUnit\Framework\TestCase
```
markrogoyski, line 21: The file name and class name do not match. I suggest changing to `QuaternionAxiomsTest` to match the file name.

FYI, PHPUnit gives this warning:
```
Warning:       Test case class not matching filename is deprecated
               in /tests/Number/QuaternionAxiomsTest.php
               Class name was 'QuaternionNumberAxiomsTest', expected 'QuaternionAxiomsTest'
```

src/Number/Quaternion.php
```diff
@@ -0,0 +1,343 @@
...
+            $k = $this->k;
+        } elseif ($q instanceof Quaternion) {
+            $r = $this->r + $q->r;
+            $i = $this->i + $q->i;
+            $j = $this->j + $q->j;
+            $k = $this->k + $q->k;
+        } else {
+            throw new Exception\IncorrectTypeException('Argument must be real or quaternion');
+        }
+        return new Quaternion($r, $i, $j, $k);
```
markrogoyski, line 207: I recommend two changes in how this (and subtract/multiply, etc.:) are structured to simplify and avoid potential for bugs: return early and avoid state.

The idea of return early is do all your bail out checks first and return from the function as soon as you know you can't proceed. It standardizes where you expect to see the error cases and allows the remaining majority of the function body to be the main logic without mixes various checks for validity.

Therefore, I'd recommend a statement like this at the very top of the method:
```php
if (!is_numeric($q) && ! $q instanceof Quaternion) {
    throw new Exception\IncorrectTypeException('Argument must be real or quaternion: ' . print_r($q, true));
}
```

The next suggestion is to avoid state. This is one of the lessons from functional programming. Keeping state around and mixing it allows the potential for bugs where the state carries over. So I would suggest returning a new object immediately for each case.

```php
        if (is_numeric($q)) {
            $r = $this->r + $q;
            return new Quaternion($r, $this->i, $this->j, $this->k);
        }

        $r = $this->r + $q->r;
        $i = $this->i + $q->i;
        $j = $this->j + $q->j;
        $k = $this->k + $q->k;

        return new Quaternion($r, $i, $j, $k);
```

Put it all together it would look like this.

```php
        if (!is_numeric($q) && ! $q instanceof Quaternion) {
            throw new Exception\IncorrectTypeException('Argument must be real or quaternion' . print_r($q, true));
        }

        if (is_numeric($q)) {
            $r = $this->r + $q;
            return new Quaternion($r, $this->i, $this->j, $this->k);
        }

        $r = $this->r + $q->r;
        $i = $this->i + $q->i;
        $j = $this->j + $q->j;
        $k = $this->k + $q->k;

        return new Quaternion($r, $i, $j, $k);
```

The numeric check is technically in there twice, but it has different purposes. The first is for an early return bail out, and the second is for logic.

src/Number/Quaternion.php
```diff
@@ -0,0 +1,343 @@
...
+            $g = $q->j;
+            $h = $q->k;
+        } else {
+            throw new Exception\IncorrectTypeException('Argument must be real or quaternion');
+        }
+        return new Quaternion(
+            $a * $e - $b * $f - $c * $g - $d * $h,
+            $b * $e + $a * $f + $c * $h - $d * $g,
+            $a * $g - $b * $h + $c * $e + $d * $f,
+            $a * $h + $b * $g - $c * $f + $d * $e
```
markrogoyski, line 275: I'm not happy with the PHPCS rule about all operators having a space around them. If you want, go ahead and add this rule to the `tests/coding_standard.xml` to allow for flexibility in how operators are written.
```
  <!-- Sometimes it is nice to write math operations like this: 3/2 -->
  <exclude name="PSR12.Operators.OperatorSpacing.NoSpaceBefore"/>
  <exclude name="PSR12.Operators.OperatorSpacing.NoSpaceAfter"/>
```

This would allow you to write this like this:
```php
        return new Quaternion(
            $a*$e - $b*$f - $c*$g - $d*$h,
            $b*$e + $a*$f + $c*$h - $d*$g,
            $a*$g - $b*$h + $c*$e + $d*$f,
            $a*$h + $b*$g - $c*$f + $d*$e
        );
```

And next time you have a fraction, you can write it like a fraction `5/2` instead of like division`5 / 2`

src/Number/Quaternion.php
```diff
@@ -0,0 +1,343 @@
...
+        $a = $this->r;
+        $b = $this->i;
+        $c = $this->j;
+        $d = $this->k;
+        if (is_numeric($q)) {
+            $e = $q;
+            $f = 0;
+            $g = 0;
+            $h = 0;
+        } elseif ($c instanceof Quaternion) {
```
markrogoyski, line 263: Should this be `$q` rather than `$c`?

src/Number/Quaternion.php
```diff
@@ -0,0 +1,340 @@
...
+
+        $e = $q->r;
+        $f = $q->i;
+        $g = $q->j;
+        $h = $q->k;
+        return new Quaternion(
+            $a*$e - $b*$f - $c*$g - $d*$h,
+            $b*$e + $a*$f + $c*$h - $d*$g,
+            $a*$g - $b*$h + $c*$e + $d*$f,
+            $a*$h + $b*$g - $c*$f + $d*$e
```
markrogoyski, line 272: Just looking at the Wikipedia definition for this, and I was wondering if you thought it might be easier to understand if you used: a₁ b₁, c₁, d₁ and   a₂, b₂, c₂, d₂.
スクリーンショット 2021-07-06 12 45 23

markrogoyski:
Looks good. Are you planning to add anything more or ready to merge?

Beakerboy:
I’m fine with it being merged. I remembered that another user had started work on this, and thought I’d finish it up to get it in.

markrogoyski:
Thanks for the PR and finishing this up.

PR #420, opened 2021-07-12
Polynomial Companion matrix

Beakerboy:


tests/Expression/PolynomialTest.php
```diff
@@ -1257,4 +1257,56 @@ public function dataProviderForNegate(): array
...
+     */
+    public function testCompanionMatrix(array $roots, array $expected_matrix)
+    {
+        // Initialize a polynomial to 1
+        $poly = new Polynomial([1]);
+
+        // Generate the polynomial with the provided roots
+        foreach ($roots as $root) {
+            $factor = new Polynomial([1, -1 * $root]);
+            $poly = $poly->multiply($factor);
```
markrogoyski, line 1275: This is too tricky to follow for a unit test. We write tests to check our logic. If we have logic in the tests, who tests the tests?

How about just putting the fully-constructed polynomial in the data provider so it is clear and obvious what the given input is that is computing the companion matrix, and it simplifies the test.

```php
    public function dataProviderForTestCompanionMatrix()
    {
        return [
            [
                new Polynomial([1, -21, 175, -735, 1624, -1764, 720]),
                [
                    [0, 0, 0, 0, 0, -720],
                    [1, 0, 0, 0, 0, 1764],
                    [0, 1, 0, 0, 0, -1624],
                    [0, 0, 1, 0, 0, 735],
                    [0, 0, 0, 1, 0, -175],
                    [0, 0, 0, 0, 1, 21],
                ],
            ],
// ...
```

Seeing the Polynomial with all its values in the constructor next to the expected companion matrix makes it much easier to see how the expected companion matrix relates to the given polynomial and whether it is correct or not.

src/Expression/Polynomial.php
```diff
@@ -440,4 +443,21 @@ public function roots(): array
                 return [\NAN];
         }
     }
+
+    /**
+     * The companion matrix for a polynomial
+     *
+     * https://en.wikipedia.org/wiki/Companion_matrix
+     */
+    public function CompanionMatrix(): NumericSquareMatrix
```
markrogoyski, line 452: lowerCamelCase on the method name: `CompanionMatrix` -> `companionMatrix`.

Beakerboy: I wonder why PHPCS didn’t catch this?

tests/Expression/PolynomialTest.php
```diff
@@ -1257,4 +1257,56 @@ public function dataProviderForNegate(): array
...
+                ],
+            ],
+            [
+                [0, -5],
+                [
+                    [0, 0],
+                    [1, -5],
+                ],
+            ],
+        ];
```
markrogoyski, line 1310: Add a test case of single constant value polynomial like `new Polynomial([5])`. Those seem to fail with `OutOfBoundsExceptions`.

src/Expression/Polynomial.php
markrogoyski: I have these rules in the coding standard which allows things like greek-letter method names:
```
  <exclude name="PSR1.Methods.CamelCapsMethodName.NotCamelCaps"/>
  <exclude name="Generic.NamingConventions.UpperCaseConstantName.ClassConstantNotUpperCase"/>
```

tests/Expression/PolynomialTest.php
Beakerboy: A constant is not a polynomial and would not have a companion matrix. I’ll probably make the new function throw an exception.

tests/Expression/PolynomialTest.php
```diff
@@ -1257,4 +1257,56 @@ public function dataProviderForNegate(): array
...
+                    [0, 0, 1, 0, 0, 735],
+                    [0, 0, 0, 1, 0, -175],
+                    [0, 0, 0, 0, 1, 21],
+                ],
+            ],
+            [
+                [6, -5],
+                [
+                    [0, 30],
+                    [1, 1],
```
markrogoyski, line 1300: For a polynomial of one degree, should the answer be a 2x2 matrix or a 1x1?

For reference, here is NumPy's answer:
```python
In [10]: p1 = [-5, 6]

In [11]: numpy.polynomial.polynomial.polycompanion(p1)
Out[11]: array([[0.83333333]])
```

And this online calculator agrees with NumPy here: https://www.easycalculation.com/matrix/companion-matrix.php

I'm not so familiar with this but can you investigate what the correct behavior should be for this?

tests/Expression/PolynomialTest.php
markrogoyski: Why does this test case work then?
```
            [
                [0, -5],
                [
                    [0, 0],
                    [1, -5],
                ],
            ],
```

tests/Expression/PolynomialTest.php
Beakerboy: The array is not a polynomial object, but a list of the roots of a particular polynomial.

markrogoyski: Actually, I forgot about the logic in the test method constructing the polynomial.

markrogoyski: This goes to my point about tricky logic in unit tests =). I already forgot how it works.

tests/Expression/PolynomialTest.php
Beakerboy: the original purpose of the companion matrix was to help us calculate the roots of higher order polynomials using the eigenvalues of the companion matrix. These tests were set up to accept an array of roots, construct the polynomial, calculate the companion matrix, then demonstrate that the roots could be recovered from the polynomial. [0,5] would mean x²+5x from (x+0) * (x+5).

Beakerboy:
I believe all your comments have been addressed.

markrogoyski:
Random question/comment not related to this PR but to the Polynomial class. The constructor currently takes coefficients ordered from high degree to low degree. That seems to make sense if you are manually manipulating code and thinking about how you might write it. But I was thinking maybe it makes more sense to order the coefficients from low to high degree, as an automated system would always have the same indexes for each degree.

markrogoyski:
Looks good. Thanks for breaking this into a smaller PR and continuing to work on it.

Any plans to continue with root finding and eigen decomposition?

Beakerboy:
It’s possible this was an implementation detail that @jakobsandberg choose. We can organize coefficients either way

markrogoyski:
It's a backwards-compatible breaking change, so it wouldn't be until the next major version if we did it. But I was using `numpy.polynomial.polynomial.polycompanion` today and it seemed to make a lot of sense to order the coefficients from low to high degree from an automated systems/programatic perspective.

Beakerboy:
> Looks good. Thanks for breaking this into a smaller PR and continuing to work on it.
> 
> Any plans to continue with root finding and eigen decomposition?

Yeah, I’d like to get that working. I think you had questions about how to handle imaginary roots.

Beakerboy:
> It's a backwards-compatible breaking change, so it wouldn't be until the next major version if we did it. But I was using `numpy.polynomial.polynomial.polycompanion` today and it seemed to make a lot of sense to order the coefficients from low to high degree from an automated systems/programatic perspective.

We could change the internal implementation at any time though.

markrogoyski:
> We could change the internal implementation at any time though.

It's the public constructor that I'm talking about changing.

PR #421, opened 2021-07-12
Svd

Beakerboy:


Beakerboy:
In the earlier discussion, you had a test for the matrix:
```
[[1,1,1,1,1]
[1,1,1,1,1]
[1,1,1,1,1]]
```
and were worried that the `getV()->isOrthogonal()` test failed. It looks like the V matrix will only be orthogonal if the input matrix is full rank. since Gaussian elimination of the source matrix will have two rows of all zeros, it is not full rank,
I could add your case as a test, and only assert that it is orthogonal if the rank equals the number of rows if you would like.

markrogoyski:
> In the earlier discussion (#338), you had a test for the matrix:
> 
> ```
> [[1,1,1,1,1]
>  [1,1,1,1,1]
>  [1,1,1,1,1]]
> ```
> 
> and were worried that the `getV()->isOrthogonal()` test failed. It looks like the V matrix will only be orthogonal if the input matrix is full rank. Since Gaussian elimination of the source matrix will have two rows of all zeros, it is not full rank,
> I could add your case as a test, and only assert that it is orthogonal if the rank equals the number of rows if you would like.

I recommend making it a separate test case and not having unnecessary logic in unit tests.

markrogoyski:
Thanks for following up with this pull request!

src/LinearAlgebra/Decomposition/SVD.php
```diff
@@ -0,0 +1,137 @@
...
+
+        $diag = $S->getDiagonalElements();
+
+        // If there is a negative singular value, we need to adjust the signs of columns in U
+        if (min($diag) < 0) {
+            $sig = MatrixFactory::identity($U->getN())->getMatrix();
+            foreach ($diag as $key => $value) {
+                $sig[$key][$key] = $value >= 0 ? 1 : -1;
+            }
+            $signature = MatrixFactory::create($sig);
```
markrogoyski, line 109: You can use `createNumeric` here which will produce the right type expectation and make static analysis tools and PHPStorm happier.

tests/LinearAlgebra/Decomposition/SVDTest.php
```diff
@@ -0,0 +1,157 @@
...
+        // Given
+        $A = MatrixFactory::create($A);
+        $S = MatrixFactory::create($expected['S']);
+
+        // When
+        $svd = $A->SVD();
+        $svdU = $svd->U;
+        $svdS = $svd->S;
+        $svdV = $svd->V;
+        $svdD = $svd->D;
```
markrogoyski, line 32: The `$svdD` variable is not used. Did you intend to run an assertion on it in this test?

tests/LinearAlgebra/Decomposition/SVDTest.php
```diff
@@ -0,0 +1,157 @@
...
+                ],
+                [
+                    'S' => [
+                        [3.872983, 0, 0, 0, 0],
+                        [0, 1.812987e-16, 0, 0, 0],
+                        [0, 0, 1.509615e-32, 0, 0],
+                    ],
+                ],
+            ]
+        ];
```
markrogoyski, line 111: This test case seems to crash.
```php
            [
                [[0]],
                [
                    'S' => [[0]],
                ],
            ],
```
```
1) MathPHP\Tests\LinearAlgebra\Decomposition\SVDTest::testSVD with data set #6 (array(array(0)), array(array(array(0))))
MathPHP\Exception\BadParameterException: Parameter λ cannot equal 0

/math-php/src/LinearAlgebra/NumericMatrix.php:1344
/math-php/src/LinearAlgebra/Eigenvector.php:48
/math-php/src/LinearAlgebra/NumericMatrix.php:2998
/math-php/src/LinearAlgebra/Decomposition/SVD.php:93
/math-php/src/LinearAlgebra/NumericMatrix.php:2789
/math-php/tests/LinearAlgebra/Decomposition/SVDTest.php:28

2) MathPHP\Tests\LinearAlgebra\Decomposition\SVDTest::testSVDProperties with data set #6 (array(array(0)), array(array(array(0))))
MathPHP\Exception\BadParameterException: Parameter λ cannot equal 0

/math-php/src/LinearAlgebra/NumericMatrix.php:1344
/math-php/src/LinearAlgebra/Eigenvector.php:48
/math-php/src/LinearAlgebra/NumericMatrix.php:2998
/math-php/src/LinearAlgebra/Decomposition/SVD.php:93
/math-php/src/LinearAlgebra/NumericMatrix.php:2789
/math-php/tests/LinearAlgebra/Decomposition/SVDTest.php:138
```

For reference, R and SciPy seem to handle it.

R
```R
> A = rbind(c(0))
> svd(A)
$d
[1] 0

$u
     [,1]
[1,]    1

$v
     [,1]
[1,]    1
```

Python
```python
In [19]: scipy.linalg.svd([[0]])
Out[19]: (array([[1.]]), array([0.]), array([[1.]]))
```

markrogoyski:
Thanks for following up with this.

I tried a lot of test cases and they all seemed to work. The only problem I encountered was the 1x1 [[0]] matrix.

Beakerboy:
Fixed comments:

* Use createNumeric()

* Fix divide by zero error for eigenvectors of zero matrix

* Separate full rank from lesser matrices in SVD Test

Beakerboy:
I believe all your comments have been addressed.

markrogoyski:
Thanks. This is great. SVD can enable a lot of other matrix functionality. Thanks for following up.

PR #422, opened 2021-07-13
Zipf's Law

Beakerboy:


src/Probability/Distribution/Discrete/Zipf.php
```diff
@@ -0,0 +1,134 @@
...
+     * @return number
+     */
+    public function pmf(int $k)
+    {
+        $s = $this->s;
+        $N = $this->N;
+
+        $n = $b - $a + 1;
+
+        return 1 / $k ** $s NonInteger::generalizedHarmonicNumber($N, $s);
```
markrogoyski, line 68: Looks like you are missing an operator here.

src/Probability/Distribution/Discrete/Zipf.php
```diff
@@ -0,0 +1,134 @@
...
+     * pmf = ---------
+     *           n
+     *
+     * Percentile n = b - a + 1
+     *
+     * @param number $k percentile
+     *
+     * @return number
+     */
+    public function cdf(int $k)
```
markrogoyski, line 84: This entire function looks like it was copied from the Uniform distribution and isn't actually implemented for Zipf. Looks like this should be some ration of harmonic numbers.

Beakerboy:
Yeah, there are a lot of issues. don't bother looking at it until the green checkmark is up.

markrogoyski:
Neat. Never heard of Zipf until now. Thanks for working on this.

Beakerboy:
I'm trying to clear out some of my older branches of new ideas. Zipfs law is related to Benford's Law.

Beakerboy:
I looks like I incorrectly named the Generalized Harmonic Series as Hyperharmonic Series some time ago. The Hyperharmonic Series is slightly different. I don't know how you want to deal with that.

markrogoyski:
> I looks like I incorrectly named the Generalized Harmonic Series as Hyperharmonic Series some time ago. The Hyperharmonic Series is slightly different. I don't know how you want to deal with that.

Sounds like it is wrong, so the fix is to fix it. And add a new function for generalized harmonic. If the method interface changes on the Hyperharmonic then I guess we just live with that breaking change. It was wrong anyways.

BTW, this would be a good time to fix the method names in that class to lowerCamelCase.

Thanks.

src/Sequence/NonInteger.php
```diff
@@ -55,10 +58,55 @@ public static function Hyperharmonic(int $n, float $p): array
...
+     * @param int $n the length of the sequence to calculate
+     * @param int $r the depth of recursion
+     * @param int $rational return results as a Rational object
+     *
+     * @return array
+     */
+    public static function hyperharmonic(int $n, int $r, $rational = false): array
+    {
+        if ($r < 0) {
+            throw new Exception\OutOfBoundsException('Recursion depth must be greater than 0');
```
markrogoyski, line 86: Exception message and condition don't match. Should the condition be `if ($r <= 0)`?

Wikipedia shows r > 0:
スクリーンショット 2021-07-13 14 25 40

src/Sequence/NonInteger.php
```diff
@@ -55,10 +58,55 @@ public static function Hyperharmonic(int $n, float $p): array
...
+                $∑ = $∑->add($array[$k]);
+                $sequence[$k] = $∑;
+            }
+        }
+        if (!$rational) {
+            for ($k = 1; $k <= $n; $k++) {
+                $sequence[$k] = $sequence[$k]->toFloat();
+            }
+        }
+        return $sequence;
```
markrogoyski, line 110: Rather than mutate and share state between conditions and a singular return value, you can return early here to simplify the logic.

```php
if ($rational == true) {
    return $sequence;
}

return array_map(
    function (Rational $r) {
        return $r->toFloat();
    },
    $sequence
}
```

src/Sequence/NonInteger.php
Beakerboy: No, the message should be “cannot be less than zero”. The special case for r=0 is what halt’s the recursion.

src/Sequence/NonInteger.php
Beakerboy: Ah yes..the ol' array_map. I always forget about it.

Beakerboy:
Both issues have been addressed.

src/Sequence/NonInteger.php
markrogoyski: Also FYI, since it is relevant, here, `array_map` does not reset array key indexes.

```php
php > $a = [1 => 1, 2, 3, 4];
php > print_r($a);
Array
(
    [1] => 1
    [2] => 2
    [3] => 3
    [4] => 4
)
php > $b = array_map(function ($x) { return $x + 5; }, $a);
php > print_r($b);
Array
(
    [1] => 6
    [2] => 7
    [3] => 8
    [4] => 9
)
```

src/Sequence/NonInteger.php
```diff
@@ -55,10 +58,59 @@ public static function Hyperharmonic(int $n, float $p): array
...
+     *
+     * https://en.wikipedia.org/wiki/Hyperharmonic_number
+     *
+     * @param int $n the length of the sequence to calculate
+     * @param int $r the depth of recursion
+     * @param int $rational return results as a Rational object
+     *
+     * @return array
+     */
+    public static function hyperharmonic(int $n, int $r, $rational = false): array
```
markrogoyski, line 83: What are reasonable upper bounds for n and r? Set high enough, and both will cause TypeErrors due to `lcm` used by Rational to overflow and produce floats instead of ints.

markrogoyski:
Do you know of an R or Python library or online calculator or table for generating test data for both the hyperharmonic and generalized harmonic number series?

src/Sequence/NonInteger.php
Beakerboy: How high can they go, roughly? I thought to use the Rational class here due to all the recursion. I was worried that floating point errors would be seriously amplified. I don’t know what reasonable values are, and I don’t know if this function is actually used by anyone. You asked that I replace it when I noticed the name was wrong. I’d be fine with just removing it.

alternatinely, I could stop using Rational and just use ints and floats…or the Rational object can be altered to use ArbitraryIntegers behind the scenes.

Beakerboy:
I do not Know of any online calculators. However, when n approaches infinity the generalized harmonic series becomes the Riemann zeta function.

src/Sequence/NonInteger.php
markrogoyski: It's fine. I'll probably just add some error handling so it will throw a controlled exception rather than a TypeError.

PR #426, opened 2021-07-14
Complex pow Fix for the 0^0 issue

Beakerboy:


markrogoyski:
Thanks. I took your solution and manually added it. The PR test had an issue with the test data.

PR #427, opened 2021-07-18
Update ArbitraryIntegerTest.php

Beakerboy:
The ArbitraryInteger class does not handle negative exponents correctly The result should be '-1', but is '1' instead.

```
// Given
$int =  new ArbitraryInteger('-1');

 // When
$pow = $int->pow('-1');

// Then
$this->assertSame('-1', (string) $pow);
```

I think it's a failure in the FullIntDiv function, this function also appears to not handle negative integers correctly.


src/Number/ArbitraryInteger.php
```diff
@@ -603,15 +621,18 @@ public function fullIntdiv($divisor): array
     public function pow($exp): ArbitraryInteger
     {
         $exp = self::create($exp);
-        if ($exp->equals(0)) {
+        if (!($this->equals(1) || $this->equals(-1)) && $exp->lessThan(0)) {
+            throw new Exception\BadParameterException('Negative exponents rarely produce integer results.');
```
markrogoyski, line 625: For this case, would it make sense to return a `Rational` number rather than bailing out? The native PHP behavior of int**-int returns float after all.
```php
php > var_dump(2**-2);
float(0.25)
```

Beakerboy: Yeah…hmmm….what if the ArbitraryInteher is greater than PHP_INT_MAX?

markrogoyski: The language has limits, with `PHP_INT_MAX` being one of them. Within the bounds of those limits reasonable operations should produce reasonable results. Outside of those bounds it is OK to throw exceptions as long as they are well defined and it makes sense why it happens.

Beakerboy: I’ve added in a Rational return path within the pow() function, when given a negative integer.

markrogoyski:
Thanks for working on this. Any other changes or ready to merge?

Beakerboy:
I think it’s ready to go.

PR #433, opened 2021-10-12
Student t cdf

Beakerboy:
- Added LogBeta
- Added LogGamma
No need to refactor Beta CDF at this time.

### DIFF
--- a/tests/Functions/SpecialTest.php
+++ b/tests/Functions/SpecialTest.php
@@ -467,6 +467,7 @@ class SpecialTest extends \PHPUnit\Framework\TestCase
             [1, \INF, -\INF],
             [1, 11, -2.397895],
             [5E-307, 5, 705.2842],
+            [94907000, 11, -186.9481],
         ];
     }
 


markrogoyski:
Cool. Thanks for adding the log beta and gamma. I see you are still adding test cases to hit all the edge cases. Let me know when you are ready to merge. Looks good based on the tests! Thanks again.

markrogoyski:
My bad. I didn't trace deep enough. It is indeed in the regularizedIncompleteBeta.

These two unit tests represent what is happening in the scenario.

```php
    public function testBugRegularizedBeta1()
    {
        // Given
        $x = 0.0041461509490402;
        $a = 0.5;
        $b = 170.5;

        // And
        $expected = 0.7657225092554765;

        // When
        $beta = Special::regularizedIncompleteBeta($x, $a, $b);

        // Then
        $this->assertEqualsWithDelta($expected, $beta, 0.00001);
    }
```

Reference SciPy
```python
In [9]: scipy.special.betainc(0.5, 170.5, 0.0041461509490402)
Out[9]: 0.7657225092554765
```

```php
    public function testBugRegularizedBeta2()
    {
        // Given
        $x = 0.0041461509490402;
        $a = 1.5;
        $b = 170.5;

        // And
        $expected = 0.29887797299850866;

        // When
        $beta = Special::regularizedIncompleteBeta($x, $a, $b);

        // Then
        $this->assertEqualsWithDelta($expected, $beta, 0.00001);
    }
```

Reference SciPy
```python
In [10]: scipy.special.betainc(1.5, 170.5, 0.0041461509490402)
Out[10]: 0.29887797299850866
```

Both tests are currently producing NAN instead of the answer.

markrogoyski:
The `iBetaCf` method is private, so I can write a unit test for it, but internally it calls the `beta` function twice and produces NANs.

```php
 beta(1.5, 170.5):NAN
 beta(1.5, 170.5):NAN
```

Unit test to directly make that call:
```php
    public function testBugBeta()
    {
        // Given
        $x = 1.5;
        $y = 170.5;

        // And
        $expected = 0.0003971962;

        // When
        $beta = Special::beta($x, $y);

        // Then
        $this->assertEqualsWithDelta($expected, $beta, 0.00001);
    }
```

R code
```R
> beta(1.5, 170.5)
[1] 0.0003971962
```

Beakerboy:
When $x is zero or a negative integer, should it be infinity, Nan, or out of bounds? It’s undefined. It currently returns infinity, but R returns Nan.

markrogoyski:
Undefined seems like the natural use case for NAN. But in this case maybe throw the NAN exception instead with the error message explaining that it is undefined?

markrogoyski:
> I think everything is covered as well as it can be. One condition is theoretically unreachable, but I left it in to be safe from a divide by zero. The other should be executed with the test: Gamma(-2E-309) = -\INF but I don’t know why that’s not showing up on coveralls.

For the Gamma(-2E-309) = -\INF test case, there is a nested if statement 
```php
if ($y <= 10) { ... if ($n < 0) {
```
And `n` is -2 at this point, so it enters the if block which returns a value.

There is a for loop which calculates the result and immediately returns it:
```php
for ($i = 0; $i < $n; $i++) {
    $value /= ($x + $i);
}
```
`value` starts off as 1 and `x` starts off as -2.0E-309.
```php
php > echo 1 / -2.0E-309;
-INF
```

The if block that isn't covered that I think you are talking about is this:
```php
                if ($y < $xsml) {
                    // ML_WARNING(ME_RANGE, "gammafn");
                    if ($x > 0) {
                        return \INF;
                    } else {
                        return -\INF;
                    }
                }
```
At that point, `y` is 1 and `xsml` is 2.2474362225599E-308, so it gets passed over, as y is larger.

Beakerboy:
Then I’m not sure this section of code is reachable.

Issue #458, opened 2022-11-23
Bug for the t-test?


leonardo-boulay:
Let's take these values:

$values = array();
for ($i=0; $i < 95; $i++) {$values[] = 1;}
for ($i=0; $i < 5; $i++) {$values[] = 0.5;}
for ($i=0; $i < 12; $i++) {$values[] = 0;}

$arraytTest = Significance::tTest($values, 0.569);

this leads to:

Warning: Division by zero in C:\xampp\htdocs\chess-boulayo\vendor\markrogoyski\math-php\src\Functions\Special.php on line 1121

Warning: Division by zero in C:\xampp\htdocs\chess-boulayo\vendor\markrogoyski\math-php\src\Functions\Special.php on line 1121

Warning: Division by zero in C:\xampp\htdocs\chess-boulayo\vendor\markrogoyski\math-php\src\Functions\Special.php on line 1121

if I do the same with tTestOneSampleFromSummaryData, but rounding numbers, it works.

markrogoyski:
Hi @leonardo-boulay,

Thank you for your interest in MathPHP, and thank you for reporting this issue.

I was able to reproduce the issue using your input. It looks like the issue is deep down the stack in the `regularizedIncompleteBeta`, and specifically in the helper `iBetaCF` regularized incomplete beta function. I'll take a look and see if I can determine the cause.

Thanks again for reporting this issue.
Mark

leonardo-boulay:
Thank you for your work!

leonardo-boulay:
(If you need another example, I found it)

$values = array();
for ($i=0; $i < 107; $i++) {$values[] = 1;}
for ($i=0; $i < 7; $i++) {$values[] = 0.5;}

$arraytTest = Significance::tTest($values, 0.85);

leonardo-boulay:
I temporary solved in this (horrible) way (but it works)


    private static function iBetaCF(int $m, float $x, float $a, float $b): float
    {
        $limits = [
        'x'  => '[0, 1]',
        'a'  => '(1,∞)',
        'b'  => '(1,∞)',
        ];
        $fraction_array = [];
$divisionbyzero = 1;
$counter = 0;
while ($divisionbyzero ==1 && $counter<50){
	$counter++;
        Support::checkLimits($limits, ['x' => $x, 'a' => $a, 'b' => $b]);

        $beta     = self::beta($a, $b);
        $constant = $x ** $a * (1 - $x) ** $b / $beta;
	
[...]

PR #460, opened 2022-12-27
Fix issue #458

robsheldon:
Hi,

I believe this resolves issue #458. I found a reference implementation that uses a different approach for calculating the continued fraction. This approach passes all tests for `Special::regularizedIncompleteBeta()` without triggering the division-by-zero error. It should also execute a bit faster (but I didn't run before-and-after timings).

I verified that the issue reporter's examples all ran successfully but I don't have the tools to ensure the answer is correct.

I removed the private function `Special::iBetaCF()` because it was only used by the previous version of `Special::regularizedIncompleteBeta()`.

I also added in test values extracted from the issue reporter's examples, and I double-checked the existing test values against Wolfram Alpha and adjusted a couple of them (some had a peculiar precision error (?)).

I have tried to copy math-php's style but feel free to ask me to rename anything, or adjust any syntax or formatting, or do so yourself as you see fit.


robsheldon:
I noticed there were some phpunit errors across multiple testsuites when I first started working on this. Skimming them, it looked like most were just precision errors, and math-php already has a test function to handle that. If this PR works out, I'd like to offer to spin through and resolve a bunch of the failing tests too. :-)

markrogoyski:
Hi @robsheldon,

Thank you for your pull request to address the issue with the regularized incomplete beta function. The help is appreciated!

I ran the tests on your branch locally on my Intel Mac that I've used to develop MathPHP for the past six years and all the tests passed. To troubleshoot what is going on in the PR CI tests, I pushed a trivial change in a new branch off the develop branch and I got the same test failures in the Github Actions CI due to minor floating point precision differences.

My guess is Github recently changed the containers that it runs the Github Actions in, and it has a slightly different architecture for floating point calculations. I'll look into what needs to be changed to fix this. Most likely it is just changing some `assertEquals` and `assertSame` to instead use `assertEqualsWithDelta` and provide an error precision to allow enough variability in architecture to get the right answer within some small precision difference and still pass.

Sorry for the inconvenience. I haven't pushed any changes in a few months so I had not run into this issue yet. I'll work on getting the existing tests to pass and let you know when it is fixed so your PR can get a better signal from the CI automation.

And thanks again for your pull request!
Mark

robsheldon:
Awesome! FWIW, I had a similar series of phpunit test failures in my local dev environment (php 8.1.12, debian buster), and I also got the impression that `assertEqualsWithDelta` would resolve most (all?) of them.

markrogoyski:
Hi @robsheldon,

I made some changes in the `develop` branch to fix the precision issues in the unit tests. Please merge the latest `develop` into your branch and then push that change to your pull request and I think you should have better luck with the tests.

Sorry for the trouble. And thanks again.
Mark

robsheldon:
Done! And phpunit runs all testsuites without failures in my local dev environment now. (Well ... mostly all. There was 1 failure. Might be a local Xdebug config thing. I'll investigate later and open an Issue if needed.)

Nice work. :-)

markrogoyski:
Thank you again for the pull request and for addressing the `regularizedIncompleteBeta` function by using a more stable algorithm. This will go out in the next release of MathPHP.

robsheldon:
It was genuinely my pleasure!

PR #464, opened 2023-01-26
Set::intersectPartial() method added to the SetTheory namespace

Smoren:
Hi @markrogoyski,

This is a PR about the M-partial intersection which have been previously added to your IterTools repository.

### Definition

> An **M**-partial intersection (for **M > 0**) of **N** sets is a set elements
> in which are contained in at least **M** initial sets.

### Properties

For any **N** sets:

1. **1**-partial intersection is equivalent to the
   [union](https://en.wikipedia.org/wiki/Union_(set_theory)) of these sets.
2. **2**-partial intersection is equivalent to the
   [difference](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement) of the
   [union](https://en.wikipedia.org/wiki/Union_(set_theory)) and the
   [symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference) of these sets.
3. **N**-partial intersection is equivalent to the
   [common (complete) intersection](https://en.wikipedia.org/wiki/Intersection_(set_theory)) of these sets.
4. For any **M > N** **M**-partial intersection always equals to the
   [empty set](https://en.wikipedia.org/wiki/Empty_set).

### Example

Given: sets **A**, **B**, **C**, **D** (**N = 4**).

```php
use MathPHP\SetTheory\Set;

$a = new Set([1, 2, 3, 4, 5]);
$b = new Set([1, 2, 10, 11]);
$c = new Set([1, 2, 3, 12]);
$d = new Set([1, 4, 13, 14]);
```

#### M = 1
It is equivalent to `A ∪ B ∪ C ∪ D`.

image

```php
$r = $a->intersectPartial(1, $b, $c, $d);
// [1, 2, 3, 4, 5, 10, 11, 12, 13, 14]
```

#### M = 2
It is equivalent to `(A ∪ B ∪ C ∪ D) \ (A ∆ B ∆ C ∆ D)`.

image

```php
$r = $a->intersectPartial(2, $b, $c, $d);
// [1, 2, 3, 4]
```

#### M = 3

image

```php
$r = $a->intersectPartial(3, $b, $c, $d);
// [1, 2]
```

#### M = 4 (M = N)
It is equivalent to `A ∩ B ∩ C ∩ D`.

image

```php
$r = $a->intersectPartial(4, $b, $c, $d);
// [1]
```

#### M = 5 (M > N)
Equals to an empty set.

image

```php
$r = $a->intersectPartial(5, $b, $c, $d);
// []
```


markrogoyski:
Hi @Smoren,

Thank you for your interest in MathPHP and for your pull request.

I understand the functionality from our conversations in the IterTools PRs.

I have a question about implementation. Based on your description here:

> For any N sets:

>1-partial intersection is equivalent to the [union](https://en.wikipedia.org/wiki/Union_(set_theory)) of these sets.
>2-partial intersection is equivalent to the [difference](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement) of the [union](https://en.wikipedia.org/wiki/Union_(set_theory)) and the [symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference) of these sets.
>N-partial intersection is equivalent to the [common (complete) intersection](https://en.wikipedia.org/wiki/Intersection_(set_theory)) of these sets.
>For any M > N M-partial intersection always equals to the [empty set](https://en.wikipedia.org/wiki/Empty_set).

Since this new set operation seems to be able to be defined in terms of other set operations (union, difference, symmetric difference, and intersection), and those set operations are already defined on `Set`, would it be possible to implement the proposed `partialIntersection` in terms of the other already-existing set operations?

Also, whether or not the above is possible or not, allow me to introduce how we do unit testing for MathPHP. In addition to the individual unit tests on each method (Which you added, thank you!), whenever possible, we also add axiom unit tests based on mathematical axioms that show principles that should hold regardless of how things are implemented. See the [SetAxiomsTest](https://github.com/markrogoyski/math-php/blob/master/tests/SetTheory/SetAxiomsTest.php) for examples for the class you are working on. For other examples, see [MatrixAxiomsTest](https://github.com/markrogoyski/math-php/blob/master/tests/LinearAlgebra/Matrix/Numeric/MatrixAxiomsTest.php) and [VectorAxiomsTest](https://github.com/markrogoyski/math-php/blob/master/tests/LinearAlgebra/Vector/VectorAxiomsTest.php) for instance. The definitions you stated above seem like good axiom tests to add for this new operation. For these kinds of tests, using a wide variety of inputs is recommended, as the axioms should hold regardless, and no pre-computed results generally need to be done as the axioms compute the rules that must hold.

Thanks,
Mark

Smoren:
Hi @markrogoyski,

Thanks for the feedback!

Indeed, there are cases (N=1, N=2, N=M, N>M) in which the operation is easily defined in terms of classical set operations. However, I don't see an easy way to express the partial intersection in terms of other operations for the cases 2<M<N.
The algorithm that I have implemented works universally for any M>1, without considering special cases.
Moreover, two classical operations can be considered as a special case of partial intersection:
- complete intersection (N-partial intersection of N sets for any N);
- union (1-partial intersection of N sets for any N).

As for axiomatic tests: I'll try to figure it out.

P.S. Can you unlock github workflow for me to run without manual approval?

Thank you!

Smoren:
@markrogoyski, OK, got it.

So I think this PR is ready to merge if you do not have new questions.

Smoren:
Hi @markrogoyski, I've rebase this branch from `develop` and resolved the conflicts.

PR #465, opened 2023-01-29
PHPdoc annotations added, phpstan works without errors now

Smoren:
Hi @markrogoyski,

I wrote PHPdoc annotations for library components.

Now PHPstan works without errors.

In some places, I needed to slightly modify the code (mainly typecasting).

Places that I considered controversial, I marked FIXME in the comments. Look, please, these places especially attentively.

I hope this work will be useful for the library and its users.


markrogoyski:
Hi @Smoren,

Wow, thank you for the PR and adding meta data annotations for types and such. This was a long term goal of mine to get the static analysis suggestions implemented from PHPStan.

This is a large PR with a lot of changes, so give me some time to review it. Glancing at it it looks like good stuff.

Just curious, what version of PHP and version of PHPStan did you use for testing?

Thanks again!
Mark

Smoren:
Hi @markrogoyski,

PHPstan — 1.9.4.

PHP:
```
PHP 7.4.33 (cli) (built: Jan 13 2023 10:42:56) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.33, Copyright (c), by Zend Technologies
    with Xdebug v3.1.6, Copyright (c) 2002-2022, by Derick Rethans
```

Smoren:
BTW, also pay attention to the places where I had to use `@phpstan-ignore-next-line`. Maybe you can figure out how to avoid using it in some places.

Smoren:
Hi @markrogoyski,

This branch is rebased from `develop`.

Smoren:
Hi @markrogoyski,

This branch is rebased from `develop` and added tasks for static analisys to workflows. Fixed several bugs for PHP 8+.

Now PHPstan succeed for PHP versions from 7.2 to 8.2.

src/Algebra.php
```diff
@@ -48,7 +48,7 @@ public static function gcd(int $a, int $b): int
      * @param  int $a
      * @param  int $b
      *
-     * @return array [gcd, a', b']
+     * @return array{float, float, float} [gcd, a', b']
```
markrogoyski, line 51: I think these are guaranteed to always be ints.

src/LinearAlgebra/ComplexMatrix.php
```diff
@@ -37,7 +39,7 @@ protected function validateComplexData(array $A)
     /**
      * Zero value: [[0 + 0i]]
      *
-     * @return ComplexMatrix
+     * @return ObjectMatrix
```
markrogoyski, line 42: I think the right thing to do here is for the method to return `new ComplexMatrix` instead, rather than the parent type, and then the return type annotation would also match.

src/LinearAlgebra/Eigenvalue.php
```diff
@@ -106,10 +108,16 @@ public static function closedFormPolynomialRootMethod(NumericMatrix $A): array
...
         // Calculate the roots of the determinant.
+        /** @var array<number> $eigenvalues */
         $eigenvalues = $det->roots();
         \usort($eigenvalues, function ($a, $b) {
+            /**
+             * @var number $a
+             * @var number $b
+             */
             return \abs($b) <=> \abs($a);
         });
```
markrogoyski, line 119: That is a lot of vertical space to use up within an anonymous function. I wonder if something like this would work to still pass the static analysis requirements and meet the style constraints:

```php
        \usort($eigenvalues, function ($a, $b) {
            /** @var number $a */ /** @var number $b */
            return \abs($b) <=> \abs($a);
        });
```

src/LinearAlgebra/Matrix.php
```diff
@@ -342,6 +353,8 @@ public function augment(Matrix $B): Matrix
             $⟮A∣B⟯[$i] = \array_merge($A[$i], $B[$i]);
         }
 
+        /** @var array<array<float|int|object>> $⟮A∣B⟯ */
+        /** @var static */
         return MatrixFactory::create($⟮A∣B⟯, $this->ε);
```
markrogoyski, line 358: What does `@var static` refer to here? `$ε` already has a doc block where declared.

Smoren: This token means the return value is Matrix. PHPstan swears if you do not specify it.

src/LinearAlgebra/Eigenvalue.php
Smoren: Done.

src/LinearAlgebra/ComplexMatrix.php
Smoren: Done.

src/Algebra.php
Smoren: Done.

src/NumericalAnalysis/NumericalDifferentiation/NumericalDifferentiation.php
```diff
@@ -143,12 +153,14 @@ protected static function sort(array $points): array
      * Ensures that the length of each subinterval is equal, or equivalently,
      * that the spacing between each point is equal
      *
-     * @param  array[] $sorted Points sorted by (increasing) x-component
+     * @param  array<array{int|float, int|float}> $sorted Points sorted by (increasing) x-component
      *
      * @throws Exception\BadDataException if the spacing between any two points is not equal
      *         to the average spacing between every point
+     *
+     * FIXME: maybe rename to checkIsSpacingConstant?
```
markrogoyski, line 161: Note to self: Probably rename these to `assertIsSpacingConstant` and similarly for other void functions in these classes that throw exceptions if preconditions not met.

src/SampleData/PlantGrowth.php
```diff
@@ -94,6 +95,10 @@ public function getWeight(): array
      */
     public function getGroup(): array
     {
+        /**
+         * FIXME: looks like a mistake: column_key = 0 instead of 1?
```
markrogoyski, line 99: Nice catch. These column indexes are mistakes. I'll fix them. Thanks.

src/Search.php
```diff
@@ -215,16 +226,18 @@ function ($value) {
...
+    private static function baseArgMin(array $values): ?int
     {
         $max = \min($values);
         foreach ($values as $i => $v) {
             if ($v === $max) {
                 return $i;
             }
         }
+
+        return null;
```
markrogoyski, line 240: Note: I think it makes more sense to throw an exception here as a precondition gone wrong rather than change the return type to nullable.

src/Sequence/Advanced.php
```diff
@@ -448,6 +448,11 @@ public static function magicSquares(int $n): array
         return $M;
     }
 
+    /**
+     * FIXME: big numbers are float, so accuracy is lost.
+     * php > var_dump(191561942608236107294793378084303638130997321548169216);
+     * float(1.9156194260824E+53)
+     */
     private const PERFECT_NUMBERS = [
         6, 28, 496, 8128, 33550336, 8589869056, 137438691328, 2305843008139952128, 2658455991569831744654692615953842176, 191561942608236107294793378084303638130997321548169216
```
markrogoyski, line 457: Nothing to fix here. PHP has limited integer precision.

```php
php > echo \PHP_INT_MAX;
9223372036854775807
php > echo \PHP_INT_MAX + 1;
9.2233720368548E+18
```

src/Statistics/Multivariate/PLS.php
```diff
@@ -112,10 +125,15 @@ public function __construct(Matrix $X, Matrix $Y, int $ncomp, bool $scale = fals
             $F = $F->subtract($t->multiply($c->transpose())->scalarMultiply($d));
 
             // Add each of these columns to the overall matrices
+            // @phpstan-ignore-next-line (Call to function is_null() with MathPHP\LinearAlgebra\NumericMatrix will always evaluate to false.)
             $this->C = is_null($this->C) ? $c : $this->C->augment($c);
+            // @phpstan-ignore-next-line (Call to function is_null() with MathPHP\LinearAlgebra\NumericMatrix will always evaluate to false.)
             $this->P = is_null($this->P) ? $p : $this->P->augment($p);
```
markrogoyski, line 131: Note: These values are always going to be null on the first iteration of the loop. I think the answer here is to use local variables and then set the class properties at the end of the calculation to avoid the phpstan confusion here.

src/Statistics/Regression/Methods/WeightedLeastSquares.php
```diff
@@ -57,11 +57,13 @@ public function leastSquares(array $ys, array $xs, array $ws, int $order = 1): N
     {
         // y = Xa
         $X = MatrixFactory::vandermonde($xs, $order + 1);
+        /** @var NumericMatrix $y */
         $y = MatrixFactory::createFromColumnVector($ys);
         $W = MatrixFactory::diagonal($ws);
 
         // a = (XᵀWX)⁻¹XᵀWy
         $Xᵀ       = $X->transpose();
+        // @phpstan-ignore-next-line (Call to an undefined method MathPHP\LinearAlgebra\Matrix::multiply())
```
markrogoyski, line 66: `$Xᵀ` is definitely a `NumericMatrix` here so wouldn't adding a `/** @var NumericMatrix $Xᵀ */` be the right thing to do here?

markrogoyski:
@Smoren,

Thank you for your hard work adding all the type annotations and comments to provide better static analysis during development!

Smoren:
Thank you @markrogoyski for merging this PR!

src/Sequence/Advanced.php
Smoren: OK

src/Search.php
Smoren: I think so too.

PR #473, opened 2023-11-06
Replace array_search with floating-point filter in `Eigenvector::eigenvectors`

Aweptimum:
Addresses the error found in #472 

When given eigenvalues containing duplicates, the `Eigenvector::eigenvectors` method fails in cases where the `$eigenvalues` array is made up of floats. The culprit is the floating point comparison in `array_search` when checking for orthogonal eigenvectors. Even in the case it does find a match, it can still fail if the match is at index 0 because the body of the method is guarded by a check for `!$key`. If `$key = 0`, php evaluates `!0` to true because 0 is falsey, so it can still fail to return the correct eigenvectors.

We can instead loop through the found solutions, checking if the current eigenvalue has been solved with `Arithmetic::almostEqual`, and then explicitly check if `$key === false`.

I noticed that there aren't any test cases in the EigenvectorTest that would have caught this - should I add a few?

### DIFF
--- a/src/LinearAlgebra/Eigenvector.php
+++ b/src/LinearAlgebra/Eigenvector.php
@@ -2,6 +2,7 @@
 
 namespace MathPHP\LinearAlgebra;
 
+use MathPHP\Arithmetic;
 use MathPHP\Exception;
 use MathPHP\Exception\MatrixException;
 use MathPHP\Functions\Map\Single;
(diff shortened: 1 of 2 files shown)


markrogoyski:
Hi @Aweptimum,

Thank you for the separate PR to add this improvement.

Before I merge it, I wanted to ask if you had any unit tests cases to add that would pass with the fix and fail without it? If so I'd love to add those as regression protection against future changes.

Thanks,
Mark

Aweptimum:
@markrogoyski I added a test case and rearranged the commit order so it's easier to go back and forth. I basically mimicked what happened in the QR Algorithm PR that caused failures. I put the matrices with duplicate eigenvalues in another dataprovider and added their eigenvalues into the data. In the added test, I added a really small numerical imprecision to them before passing them to the eigenvectors method. I think I may have made the offset a little too small because they don't always fail

markrogoyski:
> @markrogoyski I added a test case and rearranged the commit order so it's easier to go back and forth. I basically mimicked what happened in the QR Algorithm PR that caused failures. I put the matrices with duplicate eigenvalues in another dataprovider and added their eigenvalues into the data. In the added test, I added a really small numerical imprecision to them before passing them to the eigenvectors method. I think I may have made the offset a little too small because they don't always fail.
> 
> Edit: yeah changing from 10-15 to 10-12 makes both fail

Thanks. To be safe, due to the random element in the unit test, I ran them 1,000 times and no issues. Thanks.

PR #475, opened 2023-11-08
Add failing RREF Case

Aweptimum:
As laid out in #472, this adds a case where the calculated rref is different from the expected one


markrogoyski:
Hi @Aweptimum,

Thank you for opening a ticket and adding the test case. I'll take a look.

Mark

markrogoyski:
Have you tried manually setting `setError()` to something appropriate and then computing it?

Aweptimum:
Yes, if the error is set to 10<sup>-20</sup> it passes.

markrogoyski:
I'm inclined to think this isn't an issue, and just a natural result of numerical computations that involve divisions and comparisons. As soon as you divide numbers, you never really know what they are anymore due to floating-point precision errors. How do you determine if a very small number is intended, or zero? The library uses an arbitrary error threshold that works for most scenarios, but the option is there to override that when you know you are working with really small numbers.

How about we add this test case as a special case setting the error threshold and then call this one closed? Thoughts?

Aweptimum:
I think there's something that can be done to improve the stability because the scaled up [matrix](https://www.wolframalpha.com/input?i2d=true&i=rref+%7B%7B-2828789.05765%2C0%2C0%2C0%2C0%2C0%7D%2C%7B0%2C-2828789.05765%2C0%2C0%2C0%2C1729.7%7D%2C%7B0%2C0%2C-2828789.05765%2C0%2C-1729.7%2C0%7D%2C%7B0%2C0%2C0%2C-2828791.05765%2C0%2C0%7D%2C%7B0%2C0%2C-1729.7%2C0%2C-1.05765%2C0%7D%2C%7B0%2C1729.7%2C0%2C0%2C0%2C-1.05765%7D%7D) passes. It's like the error needs to account for the relative magnitudes of the numbers in the matrix.

Beakerboy:
If all the elements are rational numbers, the RationalNumber class could be used. It would eliminate floating point errors.

Aweptimum:
After staring at it harder last week, I realized that the rref might be right because it came from a result of `A - λI` and the λ in question is in the eigenvalue array twice so it should have 2 solutions.

So yeah you can probably just close this. There's an issue elsewhere

PR #483, opened 2024-12-01
Allow sampling from uniform and categorical distributions

balping:
Now you can sample from uniform and categorical discrete distributions by calling `rand()`.

Examples:

```php
$uniform = new Uniform(10, 20);
$random = $uniform->rand();


$categorical   = new Categorical(3, ['a' => 0.2, 'b' => 0.5, 'c' => 0.3]);
$random  = $categorical->rand(); // $random is 'a' or 'b' or 'c'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1471,6 +1471,7 @@ $probabilities = ['a' => 0.3, 'b' => 0.2, 'c' => 0.5]; // probabilities for cate
 $categorical   = new Discrete\Categorical($k, $probabilities);
 $pmf_a         = $categorical->pmf('a');
 $mode          = $categorical->mode();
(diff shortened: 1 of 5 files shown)


markrogoyski:
Hi @balping,

Thank you for your interest in MathPHP, and thank you for the PR to add random functions to the Uniform and Categorical distributions.

The Uniform distribution random function looks good.

For the Categorical distribution, how about first defining a simple public `cdf` function on the distribution that returns the sum up to the point of the category selected. For order, how about just using the order of the categories as provided in the input array when constructing the distribution? The user can decide the order upon construction. Then, then `rand` function can make use of `cdf` internally. The `cdf` function itself could set a private parameter for storing all the cumulative sums.

How does that sound?

Thank you again for your PR!
Mark

PR #488, opened 2026-01-13
Add multilinear regression

insel-maz:
This PR introduces MathPHP\Statistics\Regression\Multilinear for performing multilinear regression. The change is intended to be non-breaking.

- The Regression constructor now additionally accepts a list of features (a list of x-vectors).
- MathPHP\LinearAlgebra\MatrixFactory::vandermonde() now utilizes the new MonomialExponentGenerator.

If you would like me to make any further changes, please let me know.

Thank you very much for your comprehensive library!


markrogoyski:
Hi @insel-maz,

Apologies on the delay in reviewing this. It is on my to-do list. I will try to get back to you soon with feedback.

Thanks again for the contribution.
Mark

insel-maz:
Hi, I updated this PR.

There is now a `MonomialExponentGenerator::getNumberOfTerms($k, $order)` so we can calculate the degrees of freedom beforehand.

There is now a `Polynomial` (regression) class, which is a superset of `Multilinear`.

I also added some input validation in `Regression`. Do you think `InvalidArgumentException` is appropriate here, or should I switch to `BadDataException`?

To fix PHPStan type issues, I introduced a `TResultModel` for `LeastSquares`. It can either be the simple linear model `array{m: float, b: float}` or the more general `array<string, float>`. I’m still unsure whether to stick with a continuous numbering scheme β₀, β₁, β₂, ..., βₖ for the keys, or to use the exponents instead.

I added your unit test. If you provide more tests, I’d be happy to include them. Thanks to AI, I also added tests for Cook’s distance.

